PR #871, opened 2014-12-31
Problem:  zuuid on BSDs doens`t use the native uuid functions

mosconi:
Added a block code to use native BSD uuid API calls.

Also removed a compiler waring on zuuid (use of strcpy).

### DIFF
--- a/src/zuuid.c
+++ b/src/zuuid.c
@@ -46,11 +46,24 @@ zuuid_new (void)
         assert (sizeof (uuid) == ZUUID_LEN);
         UuidCreate (&uuid);
         zuuid_set (self, (byte *) &uuid);
-#   else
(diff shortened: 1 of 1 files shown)


src/zuuid.c
```diff
@@ -123,6 +136,9 @@ zuuid_set_str (zuuid_t *self, const char *source)
             return -1;
         self->uuid [byte_nbr] = (byte) value;
     }
+    
+    self->str[ZUUID_STR_LEN-1] = '\0';
```
Asmod4n: Doesn't this have to be
```c
self->str[ZUUID_STR_LEN] = '\0';
```
?

@mosconi @hintjens

hintjens: OK, fixed it: https://github.com/zeromq/czmq/pull/872

PR #881, opened 2015-01-06
add zhashx_values

acornejo:
I am new to zmq and czqm. I was surprised to learn that there is zhashx_keys but no zhashx_values.

I needed the latter, so I implemented it. If possible I would like this included in czmq, hence the pull-request.

Code is pretty straightforward, mostly copy & pasted from zhashx_keys and modified to create a list of values instead of keys.

### DIFF
--- a/include/zhashx.h
+++ b/include/zhashx.h
@@ -84,6 +84,12 @@ CZMQ_EXPORT size_t
 //  key_destructor as destructor for the list.
 CZMQ_EXPORT zlistx_t *
     zhashx_keys (zhashx_t *self);
+
+//  Return a zlistx_t containing the values for the items in the
+//  table. Uses the duplicator to duplicate all items and sets the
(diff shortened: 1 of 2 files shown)


hintjens:
I guess the list of values is usually just a function of the list of keys.

acornejo:
That is true, but getting a list of N keys and then doing N hash lookups is a terribly inefficient way of getting the N values.

Other languages provide this API for hash containers. For example python provides methods keys() and values() on their hashes, which as their name suggests return a list of keys and values respectively.

hintjens:
Right. Sounds good.

sappo:
@acornejo I guess you forgot to push the header file!?

taotetek:
@acornejo @sappo any response on this?

acornejo:
@sappo thanks! included header file in pull-request now.

PR #886, opened 2015-01-09
Problem: zmsg, zframe destroying object even if send failed

hintjens:
Solution: if the lower-level send fails (EGAIN, EINTR, etc.) then
do not destroy the frame or message. This lets the calling app try
to send it again later.

### DIFF
--- a/src/zframe.c
+++ b/src/zframe.c
@@ -130,8 +130,8 @@ zframe_send (zframe_t **self_p, void *dest, int flags)
         zframe_t *self = *self_p;
         assert (zframe_is (self));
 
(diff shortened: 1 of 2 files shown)


xaqq:
Hello,

If sending the first frame failed because of EGAIN or EINTR it makes sense to return an error.
However, I *think* that libzmq will not block for later frames once the first one was sent.

zmqpp retries to send the frames unless it was the first one.
I believe it would make sense to do something similar.

Thoughts ?

hintjens:
I don't think we can ever get an error on frames after the first one. It's
not clearly documented. So the assumption is that an error on any frame
means libzmq sent nothing. Resending in the binding seems dangerous.

PR #895, opened 2015-01-16
Problem: lack of min/max macros is causing some friction in Malamute

hintjens:
Solution: define these in "inevitable macros" section of czmq_prelude.h.


pijyoi:
I think this might cause problems with C++ users who would normally use std::min and std::max from \<algorithm>.

Similar to a clash that occurs with the min and max that comes from including windows.h which could be avoided with defining NOMINMAX before including windows.h.

hintjens:
I've made the definitions conditional. Presumably if windows.h already defines these, they aren't conflicting with C++...

PR #909, opened 2015-01-30
Fallback method to get network address

miska:


hintjens:
I'm going to reject this patch because the zsys_hostname () method needs to be refactored, rather than extended. The code is redundant with zbeacon, which is more accurate already. I do like the approach of connecting to Google and seeing what happens. However that does not fit the current use of zsys_hostname.

Could you put this code into a comment, and make an issue, so we don't lose this?

PR #913, opened 2015-01-30
Problem: czmq is always building makecert

febeling:
It should only happen in case the argument

  ./configure --with-makecert

is given.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,7 +56,7 @@ AC_ARG_WITH([makecert],
     AS_HELP_STRING([--with-makecert], 
         [Include the program makecert in compilation and installation.]),
     [with_makecert=$withval],
-    [with_makecert=yes])
+    [with_makecert=no])
 
 AM_CONDITIONAL([WITH_MAKECERT], [test x$with_makecert != xno])
 AM_COND_IF([WITH_MAKECERT], [AC_MSG_NOTICE([WITH_MAKECERT defined])])


sappo:
Just out of curiosity what does makecert do?

febeling:
I don't know, but it's not part of the library and only an additional tool. So it makes sense to only build it on request, in my mind.

hintjens:
We perhaps need to move tools like this to a separate project. However as
long as they're in CZMQ (like zpinger in zyre, and so on), it is surprising
to have to use a config option to build them. Also building these main
programs should be quite fast. Also, we need to keep an eye on how we do
this with zproject (which CZMQ aims for).

So perhaps this change isn't the best thing.

sappo:
After looking up what makecert does I agree with you @hintjens. I'm going to revert this pull request now.

febeling:
Thing is, the option offers to build --with-makecert, and the default is also to build with makecert. This doesn't make much sense. 

I agree that everything should be build by default, and on top the option to make --without-makecert would be nice.

The deeper problem is that I get linker errors with a static library cross-compile, but only on the makecert addon. This is a case where an option to just build without addon tools would be very convenient. In the cross-compiled use only the library is needed.

I understand that the wording of the original problem statement isn't entirely accurate, as always building everything isn't a problem as such. The real problem is that there is only an option flag for something that is default anyway, but none for changing this default to something else.

hintjens:
Ack... Sorry if I misunderstood... Please send further pull requests until
it works for you.

pmienk:
Note that any `--with-PACKAGE` options by default support `--without-PACKAGE`.  Additionally, `--with-PACKAGE` is equivalent to `--with-PACKAGE=yes`, implying functionality with `--with-PACKAGE=no` which is meant to be equivalent to `--without-PACKAGE`.  This is declared by `./configure --help` as well as documentation regarding `AC_WITH_ARGS`.  I find the summary provided by [autotools mythbusters](https://autotools.io/autoconf/arguments.html) useful.  Prior to any changes, this was the case for `makecert` as well.

PR #916, opened 2015-02-01
Problem: there isn't an option to build library without makecert add-on program

febeling:
When building this library building also the makecert add-on program is the default, and an option to turn this off isn't there yet. The option to _do_ build is present (--with-makecert) but isn't necessary, as it's the default behaviour.

Change inverts this options' semantics and renames it to --without-makecert, which resolves the problem.

(This is the next iteration at zeromq/czmq#913.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,9 +53,9 @@ PKG_PROG_PKG_CONFIG
 
 # Check for makecert intent
 AC_ARG_WITH([makecert],
-    AS_HELP_STRING([--with-makecert], 
-        [Include the program makecert in compilation and installation.]),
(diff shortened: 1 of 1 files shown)


pmienk:
This change disables the ability for a user to specify `--with-makecert=no`.  Please revert.

febeling:
@pmienk Sorry for the mess I created there. And thanks for noticing and cleaning up. I was trying to help, but my run-ins with Autotools sometimes don't end well. I didn't realise options were _always_ accepting parameters, and I think I tried `without-` version without success. But I must have done something wrong there, too. The additional indication of a default value in the help text you added is helpful.

PR #920, opened 2015-02-11
Problem: VS DLL and NuGet versions not updated for 3.x.

evoskuil:
This also includes an update to generated sources that are out of sync.


src/Makemodule.am
```diff
@@ -111,14 +111,15 @@ src_libczmq_la_LDFLAGS += \
...
+endif
+
 src_libczmq_la_LIBADD = \
     ${zmq_LIBS}
 
-if ON_ANDROID
-src_libczmq_la_LIBADD += \
-    -llog
-endif
-
```
jemc: Please revert this removal - this bit is important for logging to the system service on android.

src/Makemodule.am
```diff
@@ -135,14 +136,14 @@ TESTS = src/czmq_selftest
...
-	$(LIBTOOL) --mode=execute valgrind --tool=memcheck \
+	libtool --mode=execute valgrind --tool=memcheck \
 		--leak-check=full --show-reachable=yes --error-exitcode=1 \
 		--suppressions=$(srcdir)/src/.valgrind.supp \
 		$(srcdir)/src/czmq_selftest
 
 # Run the selftest binary under gdb for debugging
 debug: src/czmq_selftest
-	$(LIBTOOL) --mode=execute gdb -q \
+	libtool --mode=execute gdb -q \
```
jemc: Not sure why this change was made - maybe by accident?  I recall that on some systems, using `$(LIBTOOL)` instead of `libtool` was important, perhaps when there is more than one libtool, or the libtool to use was in a non-standard path.

I'd suggest this be reverted unless there is a good reason.

jemc:
Ah, I see the issue - the GSL generation scripts in `model` were never changed to reflect those additions in `src/Makemodule.am`.  And we're not yet using zproject here in CZMQ.

evoskuil:
It's hard to tell what sources are authoritative when the templates differ from their output. I fixed the template for the `-llog` issue in a new pull request. Do all of the other changes need to be fixed as well?

jemc:
@evoskuil 
Yeah, I'd bet the changes to `CMakeLists.txt` should also be reverted and migrated to the templates.  The android one was my change and my fault for not doing it in the template - I should have known better.  I always forget that CZMQ has those code generation scripts because it's not using the "standard" zproject scripts yet.

evoskuil:
Correcting this is not as easy as it first seemed, because I basically have to guess at what is the intent. Some changes were applied to the templates but not regenerated into sources, and other changes were applied directly to sources. Without rolling back to the previous versions, its hard to see the history, which will make it hard for people to correct any incorrect assumption. So I'm going to roll this all back and then bring it forward again cleanly.

evoskuil:
@hintjens I don't have permission to force back, and my attempt to revert didn't help. Can you (or another maintainer) please roll back to f47a1df8729141bfa78f4fe920b739d0ea4e3ec6 ?

evoskuil:
OK, I've issued a pull request that should revert to the generated template outputs and subsequent changes. https://github.com/zeromq/czmq/pull/924

hintjens:
I'm catching up on this thread... forgive any confusion. I've never
modified any generated code, afaik.

evoskuil:
@hintjens The revert of your commit was a failed attempt to roll back **to** your commit. The five or so generated code changes appear to have accumulated over time.

hintjens:
Ah, OK. In general I distrust myself doing anything complex using git and
would usually push new commits instead of reverting old ones (revert by
hand, so to speak).

PR #923, opened 2015-02-12
Fix broken autogen step

claws:
CZMQ is broken currently because the autogen.sh step fails. This is related to the previous commit that made an Android focused change. This change resolves the issue for a standard Linux build by placing the ``+=`` item after the ``=`` item as append can only be used after then item has been declared. I've also substitued the hardcoded project name for the project name from the `project.xml`` file for consistency with the rest of the file. 

With this change I can now build czmq normally.

NOTE: I have not tested this for Android.

### DIFF
--- a/model/build-autoconf.gsl
+++ b/model/build-autoconf.gsl
@@ -70,11 +70,6 @@ src_lib$(project.name)_la_LDFLAGS += \\
     -avoid-version
 endif
 
-if ON_ANDROID
-src_libczmq_la_LIBADD += \\
-    -llog
-endif
-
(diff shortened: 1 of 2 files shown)


evoskuil:
I'm in the process of reverting the break.

evoskuil:
See: https://github.com/zeromq/czmq/pull/924

jemc:
@claws I'm closing this because @evoskuil is working on making sure these changes make it into the template.

hintjens:
Maybe it's time to finished the CZMQ / zproject switch :-)

PR #926, opened 2015-02-12
Zproject

twhittock:
Generate build files using zproject tools, required a change to zproject to support prelude file.

Also some other seemingly minor differences.

Tested generated build for autotools on MacOS X.


hintjens:
Problem: CZMQ has a custom project generation framework

Solution: use zproject and kill the custom generators

I love it! :)

jemc:
I restarted the travis builds that seemed to be failing due to unrelated issues.  I'll merge as soon as we've confirmed they pass.

src/Makemodule.am
```diff
@@ -111,28 +114,30 @@ src_libczmq_la_LDFLAGS += \
     -avoid-version
 endif
 
-src_libczmq_la_LIBADD = \
-    ${zmq_LIBS}
-
-if ON_ANDROID
-src_libczmq_la_LIBADD += \
-    -llog
```
jemc: I noticed this link to `-llog` on android got clobbered by zproject.  We need this to log to system service on android, but as far as I know, zproject does not have an abstraction that allows us to add this linked library only on this platform.  @hintjens, any thoughts?

jemc:
As I noted in the line notes, this causes the qt-android build to fail on Travis (no longer linked with `-llog`).

twhittock:
OK well, I'm happy to implement whatever you'd like to do to fix the android issue.
Some ideas:
1. Always link liblog on Android in all zproject libraries.
2. Add some sort of platform specifier in the uses declaration `<uses project="log" platform="android" />` 
3. Add a uses declaration as 2, but don't specify platform - put platform-specific information in the known projects xml file in zproject, so we just 'use' logging, and that does the appropriate thing per platform (nothing, mostly.)
4. Update zproject to include a hand-written Makemodule-custom.am.

jemc:
@twhittock 
To me, making zproject allow for a custom optional makemodule seems the most convenient and flexible to other such issues we may encounter.

twhittock:
The CI failures are hanging tests - they don't hang on my laptop, not sure if it's related or an issue with the travis VM?

jemc:
@twhittock 
We see those hangs a lot with Travis, unfortunately.  Thanks for fixing the android build!

PR #928, opened 2015-02-14
Problem: --with-makecert disappeared in shift to zproject

hintjens:
Solution: fixed in zproject, regenerated everything here. Note
that man pages are all updated too.


Makefile.am
```diff
@@ -27,7 +30,6 @@ EXTRA_DIST = \
     version.sh
 
 include $(srcdir)/src/Makemodule.am
-include $(srcdir)/src/Makemodule-local.am # Optional project-local hook
```
twhittock: Looks like the version of zproject used did not include the android patches for local make modifications.

hintjens: Ah, sorry. Fixed now.

jemc:
Travis failure looks unrelated.  I'm merging.

PR #929, opened 2015-02-15
Problem:  configuration was broken.

### LINKED ISSUE #927: Problem: --disable-shared builds broken.
The zproject integration has changed a couple of build behaviors.

I'm not sure if the static (`--disable-shared`) test builds previously worked, because they weren't built by `make`: https://travis-ci.org/libbitcoin/libbitcoin-client/jobs/50451196#L1371

But presently it appears that test builds cannot be disabled. At the same time there is a linking failure for tests: https://travis-ci.org/libbitcoin/libbitcoin-client/jobs/50570598

Also the `--with-makecert` option is no longer available.

twhittock:
Solution: Regenerate using updated zproject

Fixes #927

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -248,6 +248,7 @@ if test "x$GCC" = "xyes"; then
     CPPFLAGS="-pedantic -Werror -Wall -Wc++-compat ${CPPFLAGS}"
 fi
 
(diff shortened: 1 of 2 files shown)


twhittock:
Weird. The actual commit message has --disable-shared in backticks, which isn't rendering in github.

PR #930, opened 2015-02-20
Problem: Unable to access zhash via other language bindings

twhittock:
Solution: Generate zhash public API via zproject CLASS api


twhittock:
CI failure was due to tests hanging...

twhittock:
OK how far should I go with this? I want to do zlist, zmsg, zsock... Basically all the basic services. Is there a point past which I should not go?

hintjens:
I guess it's worth wrapping any class that is stable. I need to flag those
in CZMQ, to mark them all as stable, draft, legacy, etc.

jemc:
@twhittock 
There is an issue here to bring up that I have encountered but not yet tried to tackle.  The language bindings generated by zproject currently has limitations in generating bindings that use classes from another (dependency) project.  It happily wraps method `m` in class `X` which returns an instance of class `Y`, as long as `X` and `Y` are part of the same project BA` (and have API models).  However, if method `n` of class `X` returns an instance of class `Z`, which is part of project `A`, zproject cannot deal with it because it does not have access to the API models from project `A` while it is generating project `B`.

This means that in my project that uses CZMQ, I currently can't return a `zmsg` or similar in an exposed method.  One of my thoughts for a "stopgap" measure would be to provide support for the core CZMQ types built-in to zproject, with appropriate conversion strategies on the inputs and outputs.  For example, in Ruby, `zmsg` could be returned as an array of strings instead of a handle to a real `zmsg`.  I consider this solution to be somewhat of a "hack" though; zproject would then be coupled to assuming a given API for `zmsg` that may or may not continue to be true, and more manual maintenance work is required in zproject.

Perhaps a better approach would be to figure out how to make the API models of project `A` available to the code generators while they are generating project `B`.

I'm not sure how this would look yet, but it bears mentioning and is relevant to this topic.  Maybe you have some ideas about this.

twhittock:
@jemc Yes this was why I started wrapping CZMQ types - zyre bindings are unusable at the moment because there's no zhash, zlist or zmsg class available.
I was thinking once I got those classes exported from CZMQ I'd focus on the import problem. A couple of thoughts as to the design:

1. Leave it as it is, and provide cast operations to the user. This is pretty unpleasant, and loses vital memory management information regarding 'fresh' versus references as return values and arguments which delete the input parameter. I'm not in favour of this at all.
2. Install the (possibly preprocessed with api nodes) project.xml for each class somewhere in /usr/local - then the build system could find these, somehow (configure --prefix notwithstanding, platform independence notwithstanding)
3. Extend `uses` to specify the classes imported from that library, as forward declarations. It becomes an error to refer to a type which has not been imported explicitly from a library. The classes exist only as the equivalent of a forward declaration, but this is enough I think to rely on it in most languages. This was my first idea, and still my favourite - it makes everything explicit, and very clear from a high level who uses which facilities of what libraries. It is more restrictive, however.

PR #938, opened 2015-02-23
Documentation & project fixes

hintjens:


twhittock:
This was done with an old zproject - quite a few issues reintroduced which have been fixed in zproject in the last few days. See #939

hintjens:
Oh... I did update zproject, and must have borked that... Sorry.

PR #940, opened 2015-02-23
Problem: generated with old/wrong version of zproject

hintjens:
Solution: update all project files


twhittock:
Note this reverted your documentation change for zframe_new

hintjens:
OK, np. Will redo properly.

PR #943, opened 2015-02-27
Fix sha1 on bigendian machines.

jcajka:
Hello, I have stumbled upon failing zdigest test on bigendian machines in fedora.

http://s390.koji.fedoraproject.org/koji/buildinfo?buildID=303454
http://ppc.koji.fedoraproject.org/koji/buildinfo?buildID=296552

Code in sha1.c seems bigendian compatible, but endianess is not checked in configure. So I have added check in configure.ac. To use result of endianity check, platform.h is included in zdigest.c.

I have tested it on s390x, powerpc64(le), x86, aarch64, arm( build/tests pass) , but it still fails on s390(31bit)..., still looking in to it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -242,6 +242,7 @@ AC_TYPE_SSIZE_T
 AC_HEADER_TIME
 AC_TYPE_UINT32_T
 AC_C_VOLATILE
+AC_C_BIGENDIAN
 
 # These options are GNU compiler specific.
 if test "x$GCC" = "xyes"; then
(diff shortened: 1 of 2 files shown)


jcajka:
For the record fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1196994

jemc:
This looks like a change that will be clobbered by zproject if not migrated to there or to a local non-generated file.

jcajka:
Thanks for, quick resolution. For the record, I had no success reproducing the failure on s390. So It seems to be probably caused by my changes to packaging/build scripts in fedora... It builds now(in fedora) without any issue.

PR #945, opened 2015-03-01
Problem: Convention in zproject for excluding implicit methods changed

jemc:
Solution: Use exclude method attribute instead of has_print class attribute


jemc:
CI failures are unrelated - we really need to get those lurking issues cleaned up...

Issue #954, opened 2015-03-13
Wrong CRT in msvc-10.0 build


dkrikun:
Just spotted out that the `Runtime Library` setting is set to multi-threaded *debug* dll, guess its a bug?

hintjens:
It's generated from zeromq/zproject, so that's the place to check.

evoskuil:
It would be a bug if you don't have a debug build selected and if you haven't explicitly selected its use.

evoskuil:
I don't see a problem.

dkrikun:
https://github.com/zeromq/czmq/blob/master/builds/msvc/properties/ReleaseDEXE.props

This shows debug runtime.

evoskuil:
Yes, that's wrong. Has been for a long time, not the fault of zproject.

jemc:
Closed by #955

PR #955, opened 2015-03-14
Problem: wrong default VC rt lib for release dynamic exe.

evoskuil:
See https://github.com/zeromq/czmq/issues/954


jemc:
I don't see a "THIS FILE IS GENERATED BY ZPROJECT" warning header at the top of this file, but it looks like something that *should* be generated by zproject.  Perhaps someone should take a look at this, so that we can count on the build being correct in all projects.

evoskuil:
These files are fixed reference data, there's no need to generate them.

PR #956, opened 2015-03-15
Problem: Unable to easily watch a directory for changes.

### LINKED ISSUE #826: Problem: Need an actor-based API for monitoring directories for changes
My applications need a way to monitor file directories in an efficient, cross-platform way.  I suspect that many other applications as well as CZMQ itself would benefit from such a feature.

CZMQ's `zcertstore` wants this feature:
https://github.com/zeromq/czmq/blob/c5f3eebcd6efff77abc95274ae6457fb7b9047fb/src/zcertstore.c#L44-L45

CZMQ's `zdir` would be able to have an API that integrates with such a feature.

Drops' `drops_agent` wants this feature:
https://github.com/edgenet/drops/blob/master/src/drops_agent.c#L50-L52

**Suggested solution:** Create a new zactor-based class ([like zbeacon](https://github.com/zeromq/czmq/blob/master/include/zbeacon.h)) that uses [inotify on unix](http://man7.org/linux/man-pages/man7/inotify.7.html) and [FindFirstChangeNotification on windows](http://msdn.microsoft.com/en-us/library/aa364417%28VS.85%29.aspx) to poll for changes in zero or more recursive directories.  It would use its zactor-pipe for control messages and use a PUB socket for change notifications.  This way, multiple actors in an application could be listening for changes while prefix filtering to hear changes for only the directories they care about.  The change messages published by the zactor would be normalized to be platform-agnostic.

I'd like this to become a part of czmq, but if its deemed too experimental to go in czmq right now, it can be developed in the zlabs.  Either way, I'll certainly develop and test it locally and make sure it is at least fairly stable before trying to get a patch merged.

twhittock:
Solution: zdir_watch actor

Starts to fix #826

### DIFF
--- a/api/zdir.xml
+++ b/api/zdir.xml
@@ -85,4 +85,35 @@
         Print contents of directory to stdout
         <argument name="indent" type="integer" />
     </method>
+
+    <method name="watch" singleton="1">
+        Create a new zdir_watch actor instance:
+
+            zactor_t *watch = zactor_new (zdir_watch, NULL);
+
+        Destroy zdir_watch instance:
+
+            zactor_destroy (&amp;watch);
+
+        Enable verbose logging of commands and activity:
+
+            zstr_send (watch, "VERBOSE");
+
+        Subscribe to changes to a directory path:
+
+            zsock_send (watch, "ss", "SUBSCRIBE", "directory_path");
+
+        Unsubscribe from changes to a directory path:
+
+            zsock_send (watch, "ss", "UNSUBSCRIBE", "directory_path");
+
+        Receive directory changes:
+            zsock_recv (watch, "sp", &amp;path, &amp;patches);
+
+            // Delete the received data.
+            free (path);
+            zlist_destroy (&amp;patches);
+        <argument name="pipe" type="zsock" />
+        <argument name="unused" type="anything" />
+    </method>
 </class>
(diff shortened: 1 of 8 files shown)


jemc:
Looks like some errors in the CI run violating `-Werror=c++-compat`.  Those should be fixed before merging.

Also, as a code style nitpick, I notice some of the `if`/`else` braces/whitespace don't match the CLASS RFC style (see Code Style : Statements in http://rfc.zeromq.org/spec:21).

twhittock:
I have no idea why this is different between my Mac and the Travis environment. The issues jemc raised are resolved in the other 2 commits, but the test failures make no sense to me. Is there a way to turn on verbose logging in travis?

jemc:
I'll try restarting the build and we can try to see if the failures are deterministic.

If it is deterministic, the only difference between builds is that succeeding one is using libzmq master, and the other three are using past stable releases.  So, try reproducing with one of those stable releases of libzmq on your dev machine.

jemc:
The libzmq master build also failed this time, so it's nondeterministic.  It's probably a race in the test or in the actor itself.

twhittock:
It was always working 100% of the time for me - the problem was something to do with running on the travis machine - maybe it was using that directory for other purposes at the same time.

By making a test-specific directory and watching that, everything went ok.

Unfortunately, the tests failed or hung for other reasons.

jossgray:
Tests pass for me as well on OSX 10.10.2

jemc:
Yep, the CI failures are now unrelated - Thanks!

PR #960, opened 2015-03-18
Problem: Unable to access zhashx via other language bindings

jossgray:
Solution: Generate zhashx public API via zproject CLASS api


jossgray:
Looks like i might have used an out of date version of zproject

include/zhashx.h
```diff
@@ -30,43 +30,63 @@ typedef void *(czmq_duplicator) (const void *item);
 typedef int (czmq_comparator) (const void *item1, const void *item2);
```
jemc: Can you remove these redundant typedefs from the top of this header?

jossgray: yup

jemc:
@jossgray 
Sorry to be confusing - I meant the other way around - can you put them back in `api/zhashx.xml`, but remove them from the preamble of the `include/zhashx.h` header?  The part of the header before the `@interface` comment is not auto-generated.

jossgray:
ah ok i see, yes.

jemc:
Never mind - I see what's going on.  What you did is okay.  If someone wants these accessible though, they'll have to add them back to a new file: `api/czmq.xml`.

twhittock:
err sorry - this is my hack from 5ea2e81 - I really don't know where they should go.

twhittock:
So merging this broke `make check-py`

jemc:
@twhittock 
If the python bindings have builtin tests, it might be worth adding another build type to https://github.com/zeromq/czmq/blob/master/ci_build.sh so that we can see more readily when something breaks them.

hintjens:
These should go in czmq_library.h, I think, as they're common to a couple
of classes.

twhittock:
I don't think there is a way to do that with zproject?

hintjens:
Sorry, I meant czmq.h (czmq_library.h contains all the generated
definitions).

We can put what we like in czmq.h, by hand.

twhittock:
Right fair enough, that's where it should go right now, but unless these are defined in a CLASS xml file, bindings won't be able to see the signature, so we won't be able to do anything with them using that stuff.

twhittock:
I think the solution is to not share this sort of thing in zproject classes - just have a new typedef per class which needs it.

PR #961, opened 2015-03-19
Problem: python bindings are not checked by CI server.

twhittock:
Solution: Add check-py target to CI build script

Due to #960 issues


jemc:
@twhittock 
It seems like python missing some part of its setup (see the error message about failing to import libc).

Also, I'd prefer to see this as a separate item in the build matrix, so it's clear when the C build is succeeding and passing, but the python bindings are failing.  If you're unsure of how to do this, I can take care of it in a separate PR after you get the python check passing in your PR.

twhittock:
ok looks like this will work now - closing and creating a PR which is readable. thanks for bearing with me here.

PR #964, opened 2015-03-20
Problem: No way to run verbose tests via make

twhittock:
Solution: Make check-verbose.

Rationalle: Newer versions of automake >= 1.13 force `make check` to be
run through a parallel test runner which turns off all output. So we
need a way to at least see the non-verbose output in that sort of
environment, hence check-local.

check-verbose comes about because when CI failures or hangs occur, there
is very little information to go on, so making it always dump as much
information as possible should hopefully help diagnose these issues in
future.

### DIFF
--- a/ci_build.sh
+++ b/ci_build.sh
@@ -14,7 +14,7 @@ if [ $BUILD_TYPE == "default" ]; then
         make check && sudo make install && sudo ldconfig ) || exit 1
 
     # Build, check, and install CZMQ from local source
-    ./autogen.sh && ./configure && make check && sudo make install
+    ./autogen.sh && ./configure && make check-verbose VERBOSE=1 && sudo make install
 else
     cd ./builds/${BUILD_TYPE} && ./ci_build.sh
 fi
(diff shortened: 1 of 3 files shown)


jemc:
Awesome, thanks!

hintjens:
I already made the animate option for this.  Make animate.

hintjens:
Sorry, that's doing something a little different, runs the tests with
verbose argument. However it overlaps with this change...

twhittock:
Yeah that was a bad commit message, really - sorry. I had seen animate before, but keep forgetting it exists, what's the art with the name? But as you say it's a subtly different rule.

Any chance someone can kick of these tests again, as I'd like to see where the tests hang.

jemc:
@twhittock 
I bumped the tests as you asked.

@hintjens 
Maybe you can give @twhittock's account access to merge PRs and restart the CI build like I have?  I don't seem to have authorization to add users the group.

twhittock:
oh it can't re-run it because it's already been merged. I guess I'll make another PR tonight and see whether we can gather information about what's going on.

hintjens:
OK, I've added @twhittock to the maintainers' team.

jemc:
@twhittock 
We have verbose captured logs of the hang here:
https://travis-ci.org/zeromq/czmq/jobs/55116189

PR #970, opened 2015-03-27
Code cleanups

hintjens:
Broke/fixed zuuid_formatted_str

### DIFF
--- a/src/zuuid.c
+++ b/src/zuuid.c
@@ -29,6 +29,7 @@
 struct _zuuid_t {
     byte uuid [ZUUID_LEN];              //  Binary UUID
     char str [ZUUID_STR_LEN + 1];       //  Printable UUID
+    char *str_canonical;                //  Canonical UUID, if any
 };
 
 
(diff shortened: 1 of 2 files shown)


twhittock:
could the function return const char * to make it clear it's not mutable?

src/zuuid.c
```diff
@@ -116,38 +118,40 @@ zuuid_set (zuuid_t *self, byte *source)
...
-            return -1;
-        uint value;
-        if (sscanf (p, "%02x", &value) != 1)
-            return -1;
-        self->uuid [byte_nbr] = (byte)value;
-        self->str[byte_nbr * 2] = *p++;
-        self->str[byte_nbr * 2 + 1] = *p++;
-        str_len -= 2;
     }
+    free (self->str_canonical);
```
twhittock: should set null here?

hintjens: OK, set to null and return 'const' in both zuuid_str and
zuuid_str_canonical.

PR #972, opened 2015-03-28
Problem: make check runs the tests twice

### LINKED ISSUE #969: Problem: "make check" runs self tests twice

twhittock:
Solution: Regenerate from zproject - project selftest not added to the
TESTS which are run by the parallel test runner (where stdout is
ignored)

Fixes #969


hintjens:
Thanks for this fast fix. :-)

PR #975, opened 2015-03-30
reversed free order can lead to crash. fix it.

bkmit:


sappo:
@bkmit thanks for the fix! For the next time it would be nice if you could state the problem you were having and explain the solution. It just makes the commit history and reviewing so much easier ;-)

For example:
```
Problem: reversed free order can lead to crash
Solution: free member before freeing struct to avoid this issue
```

bkmit:
@sapo Thank for quick merge. I will do proper commenting in next patch.

PR #981, opened 2015-04-01
Problem: There is no easy way to check if two list items are equal

sappo:
Solution: Introduce an equals function to the list which can be used by _exists and _remove methods.

I didn't choose the compare method for this problem as it doesn't work for equality. Of course I could have fix the compare method to work like strcmp for example but this would have broken user space.

### DIFF
--- a/api/zlist.xml
+++ b/api/zlist.xml
@@ -10,6 +10,13 @@
         <return type = "boolean" />
     </callback_type>
 
+    <callback_type name = "equals_fn">
+        Equals function
+        <argument name = "item1" type = "anything" />
+        <argument name = "item2" type = "anything" />
+        <return type = "boolean" />
+    </callback_type>
+
     <callback_type name = "free_fn">
         Callback function for zlist_freefn method
         <argument name = "data" type = "anything" />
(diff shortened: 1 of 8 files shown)


hintjens:
For naming, I'd prefer _eq, unless it seems weird.

sappo:
I like _eq :) and it also fits with existing stuff like streq.

sappo:
On second thought there is no equals method and zlist_eqfn () instead of zlist_equalsfn () does seem weird.

hintjens:
You don't need a new hook, though... The compare fn will do this imo.

sappo:
The current compare function return a boolean, true if item1 > item2, false otherwise. If we change it to return an integer value like strcmp () we'll break user space. Therefore I introduced the equals function.

hintjens:
That documentation is then wrong (I can't even find where it states this).
The comparator function is meant to be compatible with strcmp (indeed.
zhashx uses strcmp by default).

sappo:
You're right there's no documentation about the return value but if you have a look at zlist.h the return type is boolean and thus it is not compatible with strcmp. Should we change that?

hintjens:
Yes, let's fix this.

PR #991, opened 2015-04-20
Problem:zlist_sort function incorrect

Riddance:
Problem:zlist_sort function wont work in some case (zlist: "b", "a", "c" ) due to incorrect implement of the comb sort algorithm.
Solution:fix up the comb sort algorithm in zlist_sort function.


jemc:
Note the travis failure: 
```
src/zlist.c: In function ‘zlist_sort’:
src/zlist.c:404:14: error: variable ‘swapped’ set but not used [-Werror=unused-but-set-variable]
```

Riddance:
will fix it

PR #992, opened 2015-04-20
Problem:variable 'swapped' set but not used

Riddance:
fixed error: variable 'swapped' set but not used

### DIFF
--- a/src/zlist.c
+++ b/src/zlist.c
@@ -394,7 +394,7 @@ zlist_sort (zlist_t *self, zlist_compare_fn *compare)
     bool swapped = false;
     while (gap > 1 || swapped) {
         if (gap > 1)
-            gap = (int) ((double) gap / 1.3);
+          gap = (int) ((double) gap / 1.3);
         node_t *base = self->head;
         node_t *test = self->head;
         int jump = gap;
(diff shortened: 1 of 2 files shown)


Riddance:
fixed error: variable 'swapped' set but not used

src/zlist.c
```diff
@@ -394,14 +394,14 @@ zlist_sort (zlist_t *self, zlist_compare_fn *compare)
     bool swapped = false;
     while (gap > 1 || swapped) {
         if (gap > 1)
-          gap = (int) ((double) gap / 1.3);
+            gap = (int) ((double) gap / 1.3);
```
jemc: Can you revert this part of the change?  @hintjens just sent in [a PR to normalize the whitespace](https://github.com/zeromq/czmq/pull/990), and he shouldn't have to do it again for the same line within a few hours :wink:.

See http://rfc.zeromq.org/spec:21 for applicable style guidelines.

Riddance: sorry, i reverted it

hintjens:
@Riddance thanks for fixing zlist_sort. This is one of my favorite methods. :-)

PR #997, opened 2015-04-30
zsys_vprintf was needlessly using zmalloc

Asmod4n:
Changed zmalloc to malloc for string allocation.


Asmod4n:
The failure is unrelated to this PR.

PR #1001, opened 2015-05-04
Problem: compiler warning on zgossip_msg.c:442

### LINKED ISSUE #1000: Problem: zgossip stops czmq from compiling
```bash
src/zgossip_msg.c:442:23: error: address of array 'self->key' will always evaluate to 'true' [-Werror,-Wpointer-bool-conversion]
            if (self->key)
```

hintjens:
Solution: caused by error in zproto_codec_c; fixed in zproto and
regenerated here.

Fixes #1000

### DIFF
--- a/src/zgossip_engine.inc
+++ b/src/zgossip_engine.inc
@@ -738,7 +738,7 @@ s_server_config_service (s_server_t *self)
             if (zsock_bind (self->router, "%s", endpoint) == -1)
                 zsys_warning ("could not bind to %s (%s)", endpoint, zmq_strerror (zmq_errno ()));
         }
-#if (ZMQ_MAJOR_VERSION >= 4)
+#if (ZMQ_VERSION_MAJOR >= 4)
         else
         if (streq (zconfig_name (section), "security")) {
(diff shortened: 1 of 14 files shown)


hintjens:
@zeromq/czmq-maintainers can this pull request get some love? :)

src/zgossip_engine.inc
```diff
@@ -738,7 +738,7 @@ s_server_config_service (s_server_t *self)
             if (zsock_bind (self->router, "%s", endpoint) == -1)
                 zsys_warning ("could not bind to %s (%s)", endpoint, zmq_strerror (zmq_errno ()));
         }
-#if (ZMQ_MAJOR_VERSION >= 4)
+#if (ZMQ_VERSION_MAJOR >= 4)
```
twhittock: i think this was corrected in the source code without being pushed upstream to zproto?

twhittock: Looks like I got that the wrong way around. Never mind. Merging now.

hintjens:
Thanks!

PR #1006, opened 2015-05-15
Problem: zchunk_extend missing from header

danriegsecker:
Solution: added zchunk_extend to zchunk.h


hintjens:
I made the change and forgot to add it to the commit set... thanks for the fix!

danriegsecker:
No problem, glad to help.

PR #1013, opened 2015-05-30
Problem: documentation is out of date

hintjens:
Solution: rebuild man pages


hintjens:
Merging myself since I need this to be able to update the online man pages.,..

Issue #1014, opened 2015-06-01
czmq.dll does not include zgossip


mikegulf:
I am trying to build zyre but I get an error saying it cannot find zgossip. I used dependency walker to check czmq.dll and I found that zgossip function was missing. I checked czmq 3.0 and that one had it but I cannot use that one because zyre calls for the lastest master version.

hintjens:
The master version of CZMQ definitely builds zgossip... perhaps you're
using an older version by mistake?

mikegulf:
I went ahead and pulled down the latest version from Github this morning and tried building in MSVC 2010, 2012, and 2013 and got the same results. The versions all build but when I try to build zyre it gives me the error 

Error	1	error LNK2019: unresolved external symbol _zgossip referenced in function "void __cdecl zyre_node_gossip_start(struct _zyre_node_t *)" (?zyre_node_gossip_start@@YAXPAU_zyre_node_t@@@Z)	C:\Users\ralph\Documents\GitHub\zyre\builds\msvc\vs2012\zyre\zyre_node.obj	zyre

This is when I looked into the dll with dependency walker and found that zgossip was missing.

Thoughts?

mikegulf:
I found the issue. In zgossip.h the czmq_export was missing from in front of the zgossip and zgossip_test functions.

jemc:
@mikegulf - Note that the zgossip.h file is generated by zproto - any permanent changes will need to be fixed there.

hintjens:
OK, I'm on it...

PR #1015, opened 2015-06-01
Problem: zgossip_msg.txt should not be in doc directory

hintjens:
Solution: remove it (it's a private class).


hintjens:
Fixes #1014

PR #1018, opened 2015-06-02
Problem: zmsg_sendm not defined in API model

hintjens:
Causes loss of prototype when regenerating project.xml.

Solution: add it to right place.

Incidental fix, the Makefile also reflects zproject PR #192.


hintjens:
I'm merging this myself as it's necessary for the 3.0.1 release I'm making.

PR #1019, opened 2015-06-02
Fixes to project packaging

hintjens:


hintjens:
I can't reproduce this failure, which hits the zeromq4-x build in CZMQ.

PR #1039, opened 2015-07-07
Problem: 'makecert' conflicts with existing Mono tool

hintjens:
Solution: rename to 'zmakecert'


jemc:
Looks like the python binding CI failure is unrelated.

PR #1045, opened 2015-07-27
Link to pthread

bluca:
As reported in https://github.com/zeromq/zproject/issues/199 CZMQ uses Pthread symbols without linking to it. Although this does not cause build problems since ZMQ links to Pthread, it causes errors to be printed by the Debian build environment, since a library is expected to link against all its dependencies.

Finally, with autotool -std=gnu99 is used to build, but not with CMake, which causes a compile error since zdir uses C99's ‘for’ loop initial declarations. So, gnu99 is added to CFLAGS in the project-local CMake hooks.

Please note that this won't have any effect until https://github.com/zeromq/zproject/pull/202 is merged and ./generate.sh is run again on this project.

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -1,0 +1,7 @@
+AC_DEFUN([AX_PROJECT_LOCAL_HOOK], [
+    AC_CHECK_HEADERS(pthread.h)
+    AC_CHECK_LIB([pthread], [pthread_create],
+        [CFLAGS="${CFLAGS} -pthread"],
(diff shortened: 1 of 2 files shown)


jemc:
@bluca - I just noticed this PR (or rather, the one after it when I ran `./generate.sh` for you) broke the `qt-android` CI build.  Is there a way in the new `acinclude.m4` to exclude the `-lpthread` linkage on android platforms - or possibly make it fail softly and continue if the library is missing?

bluca:
PR incoming, sorry about that, just noticed as well on Travis.

PR #1047, opened 2015-07-27
Warning instead of error if cannot link pthread

bluca:
Android's libc, Bionic, does not provide a -pthread flag to link,
but instead links automagically when it founds <pthread.h> included.
Use AC_MSG_WARN instead of AC_MSG_ERR to avoid failing the build.

After this commit, if -pthread is not possible like on Android, the build will continue and this will be printed to warn the user:

"configure: WARNING: cannot link with -pthread."


bluca:
I will stay up and watch Travis run this time around :-)

jemc:
@bluca - Thanks for taking care of it!

bluca:
@jemc thanks for merging (and updating the build system earlier), and sorry again for the hiccup!

PR #1048, opened 2015-07-28
Add Travis target for CMake

bluca:
Hello,

Is there any particular reason for not having a CMake target in Travis?

If not, this PR enables it. As you can see from here: https://travis-ci.org/bluca/czmq/builds/72932752 it seems to work fine :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
 - BUILD_TYPE=default ZMQ_REPO=libzmq    WITH_LIBSODIUM=1
 - BUILD_TYPE=qt-android
 - BUILD_TYPE=check-py
+- BUILD_TYPE=cmake
 
 before_script:
 - sudo apt-get install uuid-dev
--- a/builds/cmake/ci_build.sh
+++ b/builds/cmake/ci_build.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Assumes Travis CI os:linux (x86_64)
+
+git clone git://github.com/jedisct1/libsodium.git &&
+( cd libsodium; ./autogen.sh && ./configure && make check && sudo make install && sudo ldconfig ) || exit 1
+
+# Build, check, and install ZeroMQ
+git clone git://github.com/zeromq/libzmq.git &&
+( cd libzmq; ./autogen.sh && ./configure && make check && sudo make install && sudo ldconfig ) || exit 1
+
+# Build, check, and install CZMQ from local source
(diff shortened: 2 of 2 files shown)


bluca:
Also in the next few days I'll try and have a look at the Python CI test, in case nobody else had time to do that yet.

jemc:
You may also want to consider adding this to the zproject's cmake build system so it's generated for all zproject-using projects.  For an example of what this looks like, see the qt-android one [https://github.com/zeromq/zproject/blob/97ae977489c14cee1c1da43c897f73ca0efa4c05/zproject_qt_android.gsl#L95-L120](here).

jemc:
Regarding the python test, it's an issue with zproject code generation for python bindings not taking certain python keywords into account and finding suitable replacements.

bluca:
Thanks, will also have a look at adding this to Zproject tomorrow or the day after.

bluca:
Unfortunately it does not look like adding CMake CI target to Zproject is possible, with the current APIs.

As far as I can see there is currently no way of making a dependency of a dependency conditional on the dependency version. Or alternatively, a dependency that only gets installed, but not added to the pkg-config check and linked against.

The problem is libsodium: when compiling CZMQ against ZMQ < 4, libsodium is not needed. When compiling against ZMQ >= 4, libsodium is needed.

By having the following in zproject_known_projects.xml:
```xml
    <use project = "zmq"
        repository = "https://github.com/zeromq/libzmq"
        test = "zmq_init"
        cmake_name = "ZeroMQ">
        <use project = "sodium" />
    </use>
```
Sodium becomes a mandatory dependency regardless of the ZMQ version, since CZMQ will always link to it.

jemc:
@bluca - Does the CI for the cmake build really need to support multiple versions of ZMQ, or can we get away with always running against `master`?

bluca:
It does run against master only, in this version.

The problem affects the main build, with autotools, but more importantly with general distribution outside of the CI: if libsodium is added as a dependency, every single build will need it, even if ZMQ < 4.

jemc:
@bluca - Can you open an issue ticket dedicated to this specific problem in the zproject repo so we can discuss there how to proceed with the issue of optional dependencies with respect to libsodium (or alternatively, file a PR there with a solution to the problem)?

bluca:
I am writing the issue as we speak :-)

bluca:
@jemc - done: https://github.com/zeromq/zproject/issues/206

PR #1051, opened 2015-07-30
Problem: Python tests failing.

twhittock:
Solution: Regenerate using latest zproject to correctly decorate methods which match a keyword

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,8 +204,6 @@ install(TARGETS czmq
     RUNTIME DESTINATION bin              # .dll file
 )
 
-include(${CMAKE_CURRENT_SOURCE_DIR}/src/CMakeLists-local.txt) # Optional project-local hook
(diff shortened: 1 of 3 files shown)


bluca:
@twhittock - thanks for fixing the Python tests.
Was there a specific reason for removing the autoconf and CMake project-local hooks? They were merged in Zproject a few days ago. If not, then I opened PR #1052 to re-generate from the latest Zproject version.

jemc:
@bluca - I'm sure it was just an accident from not having the latest zproject master when @twhittock regenerated.

twhittock:
yes purely accidental, sorry about that.

bluca:
Ah I see, no problem at all!  :-)
Thanks again for fixing the Python CI build, now the CI build is all nice and green: https://travis-ci.org/zeromq/czmq/builds/73401535

PR #1059, opened 2015-08-10
Add client server to czmq

somdoron:
* Add the socket type to zsock
* Retrieving and setting the routing id on zframe

### DIFF
--- a/include/zframe.h
+++ b/include/zframe.h
@@ -57,6 +57,11 @@ CZMQ_EXPORT zframe_t *
 CZMQ_EXPORT int
     zframe_send (zframe_t **self_p, void *dest, int flags);
 
+//  Send a reply frame to a server socket, copy the routing id from source message, destroy frame after sending.
+//  Return -1 on error, 0 on success.                     
+CZMQ_EXPORT int
+    zframe_send_reply (zframe_t **self_p, zframe_t *source_msg, void *dest, int flags);
+
 //  Return number of bytes in frame data
 CZMQ_EXPORT size_t
     zframe_size (zframe_t *self);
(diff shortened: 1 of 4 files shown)


taotetek:
Hi @somdoron !  Tests are currently failing for this PR and I don't currently have a moment to look at why - but this is a very exciting PR!  Can look this afternoon if no one else has jumped in.

somdoron:
Thanks. I will check it now...

somdoron:
so it is failing for builds that don't have server & client (zeromq 4.1,4, 3,2).

Is there a macro for zeromq version that I can use?

taotetek:
@somdoron I -think- ZMQ_VERSION_MAJOR and ZMQ_VERSION_MINOR but I could be remembering incorrectly.

taotetek:
@somdoron - aha - and zsys_version in zsys for runtime version detection

somdoron:
So I'm checking the ZEROMQ version now:

https://github.com/somdoron/czmq/commit/9342fa2cce899d187ad62bcc75e546e97b27f92c

However I'm not big fun of it, so I you have a more elegant solution let me know...

taotetek:
@somdoron if the tests pass I say let's merge it, and if someone has a more elegant solution for managing the tests that can come in a follow up PR.

I'm going to add support to goczmq for this today so I can start playing with the new types.  Fun!  I remember the conversations around this idea :)

somdoron:
Cool, I'm going to add support to zproto as well.

Asmod4n:
You can check for server and client sockets via
```c
#if defined(ZMQ_SERVER) && defined(ZMQ_CLIENT)
#endif
```

PR #1061, opened 2015-08-12
Problem: ifdefs should not check for a specific version

sappo:
Solution: Replace ugly version checking ifdefs with slightly less ugly feature checking ifdefs

Also minor style fixes.


jemc:
@sappo - I think the "minor style fixes" that remove trailing whitespace in header file comments will be overwritten by the next zproject pass, as these are due to how GSL handles blocks of text.  So, those parts of the commit should be removed and commit-amended, because they are just commit noise.  You should be able to do this by running zproject's generate and then running `git commit --amend` and the force-pushing to the branch.

sappo:
The white space issue brought to mind that changes to include file are overridden by zproject. I had to change uint32_t to size_t though as the zproject api doesn't support it.

sappo:
ping

jemc:
It looks like the Travis CI failures are unrelated, so I'll merge, but we need to figure out why there is
```
make[2]: *** No rule to make target `libzmq.la', needed by `tests/test_getsockopt_memset'.  Stop.
```
and fix it in a separate PR.  It looks like a problem in libzmq master, and not in czmq.

sappo:
I already fixed the libzmq problem a couple days ago: https://github.com/zeromq/libzmq/pull/1518

jemc:
Great!

Issue #1062, opened 2015-08-16
Ruby bindings use keywords as variables


rfestag:
I am trying to play with the Ruby bindings that are in this repository, and found that there are at least a few instances where the variable name used is actually a Ruby keyword, and thus causes syntax errors. At the top of each file, there is a warning stating that the files are generated 100% by the ZPROJECT, and shouldn't be modified, but that I should check the README for how to make permanent changes. I couldn't find any information in the README related to that. Aside from that main issue, I also have a few questions:

1) How would I go about submitting a patch for this? Is there some automatic generation of these files, or can I simply make the changes and submit a pull request?
2) Is there a reason CZMQ maintains these bindings, but does not create/maintain a gem for it? There are a few other bindings documented, but there 1-2 years old, or are native extensions (making them less portable than FFI).

Asmod4n:
You can find the api definition for the bindings here: https://github.com/zeromq/czmq/tree/master/api, the code generators for the bindings, where the changes should be made are here: https://github.com/zeromq/zproject/blob/master/zproject_bindings_ruby.gsl

The python bindings had a similar issue, the workaround was to prefix keywords: https://github.com/zeromq/zproject/blob/master/zproject_bindings_python.gsl#L116

jemc:
> Is there a reason CZMQ maintains these bindings, but does not create/maintain a gem for it?

Actually, the reason is because no one has ever put any real effort into the Ruby bindings for CZMQ yet.  There are Ruby, Python, and QML bindings generated for any C project using zproject with XML API definitions.  Other than writing the XML definitions, these bindings are generated "for free" by zproject.  Someone has put time and effort into making sure the Python bindings are usable in CZMQ, but no one has done so for the Ruby bindings for CZMQ yet.  However, I did most of the work on the Ruby bindings generator and I have been actively using the generated bindings for several other projects (just not CZMQ), so they should be sound in general, but you may run into occasional issues like the keyword related issues you mentioned.

I can make a fix today for the keyword issues you've encountered, and I'll post an update in this issue ticket when it's done.

[...]

PR #1063, opened 2015-08-16
Problem: Ruby bindings clash with self and alias keywords.

jemc:
Solution: Fix in zproject and regenerate here.

See: https://github.com/zeromq/zproject/pull/211
Relates to: https://github.com/zeromq/czmq/issues/1062


sappo:
seems like the latest libzmq tests are failing

PR #1067, opened 2015-08-24
Problem: Cannot obtain head and tail of zlistx

sappo:
Solution: Add method to get both head and tail

I also did some minor code clean-up and removed unused variable tail.


sappo:
Test failures seem unrelated

PR #1075, opened 2015-09-04
Problem: Need to forward messages based on a tokenizable route or path

sappo:
Solution: Use ztrie to write a dispatcher actor


hintjens:
Nice problem, though perhaps this belongs in a broker like Malamute rather than in CZMQ?

E.g. for service dispatching. The advantage is that replies are possible...

mhaberler:
I view  this as an extension of ztrie - where is the upside of  splitting that out? 

or do you want to split out all of ztrie? would not get my vote

sappo:
I think of zdispatcher as variant of zproxy were you can control who receives the messages or if there is anyone who cares for them. Replies from the worker are out of scope. Maybe the discuss section is little bit vague about that.

PR #1080, opened 2015-09-11
Problem: I oftentimes need to concatenate string and tend to copy the…

sappo:
Problem: I oftentimes need to concatenate strings and tend to copy the same function
Solution: Add this concatenate function to zstr

### DIFF
--- a/include/zstr.h
+++ b/include/zstr.h
@@ -64,6 +64,10 @@ CZMQ_EXPORT int
 CZMQ_EXPORT int
     zstr_recvx (void *source, char **string_p, ...);
 
+//  Concatenates a series of strings (until NULL) to a newly allocated string.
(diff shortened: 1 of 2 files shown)


sappo:
I just found zsys_sprintf which makes this pull request obsolete. 

Shouldn't this function belong into zstr instead of zsys?

hintjens:
It's in zsys because it's not portable (and we centralize non-portable code
in zsys). However it's not a clean API, I agree.

Also, zsys is a mix of external and internal (non-portable) methods, it may
be useful to split these out.

We can certainly broaden the zstr class and add more general string
manipulation.

sappo:
I can send the pull request again but I'm not sure if this code is portable.

PR #1081, opened 2015-09-11
Problem: libuuid is hardcoded as optional dependency

sappo:
Solution: Use zproject's new optional dependency instead


sappo:
Sorry for the noise, I removed some trailing whitespaces in zproject which were applied to czmq once in regenerated the project files.

PR #1083, opened 2015-09-11
Fix zframe.c build breakage

bluca:
libzmq PR https://github.com/zeromq/libzmq/pull/1577 changed a public API function call from
zmq_msg_get_routing_id to zmq_msg_routing_id, so update call in
zframe.c.


hintjens:
Thanks for doing this.

PR #1085, opened 2015-09-13
Regen from zproject, fix qtandroid/CMake, add JNI

bluca:
Adds generated FindUuid.cmake and various fixes to CMake files, and
also generates new JNI bindings.

As I wrote in the Zproject PR ( https://github.com/zeromq/zproject/pull/232 ), qt-android still needs more work to be able to build with libuuid, since on the NDK some libraries (like uuid and pthread) are not shipped as separate .so, but ship only the headers, and the current autoconf configuration does not support that.

Also,I assume it is ok to commit the new generated JNI bindings from Zproject? Please note that I have not tested those!


bluca:
test_shutdown_stress and the zauth tests seem very flacky lately. I suspect if "rerun" is hit, they'll pass. Nothing in this PR touches those areas or the autoconf build system (which is the one failing, with 4.1 ZMQ).

hintjens:
The generated JNI bindings do not work. It's fine to commit them, as work
in progress...

PR #1087, opened 2015-09-15
Added authentication commands to zproxy actor.

ritchiecarroll:
Added ability to configure CURVE and PLAIN security options through new zproxy API commands for the frontend and backend sockets of the proxy.

### DIFF
--- a/include/zproxy.h
+++ b/include/zproxy.h
@@ -73,6 +73,26 @@ extern "C" {
 //      zstr_sendx (proxy, "RESUME", NULL);
 //      zsock_wait (proxy);
 //
+//  Configure an authentication domain for the "FRONTEND" or "BACKEND" proxy
+//  socket -- see zsock_set_zap_domain (). Call before binding socket:
+//
+//      zstr_sendx (proxy, "DOMAIN", "FRONTEND", "global", NULL);
+//      zsock_wait (proxy);
+//
+//  Configure PLAIN authentication for the "FRONTEND" or "BACKEND" proxy
+//  socket -- see zsock_set_plain_server (). Call before binding socket:
+//
+//      zstr_sendx (proxy, "PLAIN", "BACKEND", NULL);
+//      zsock_wait (proxy);
+//
+//  Configure CURVE authentication for the "FRONTEND" or "BACKEND" proxy
+//  socket -- see zsock_set_curve_server () -- specifying both the public and
+//  secret keys of a certificate as Z85 armored strings -- see
+//  zcert_public_txt () and zcert_secret_txt (). Call before binding socket:
+//
+//      zstr_sendx (proxy, "CURVE", "FRONTEND", public_txt, secret_txt, NULL);
+//      zsock_wait (proxy);
+//
 //  This is the zproxy constructor as a zactor_fn; the argument is a
 //  character string specifying frontend and backend socket types as two
 //  uppercase strings separated by a hyphen:
--- a/src/zproxy.c
+++ b/src/zproxy.c
@@ -24,13 +24,14 @@
 
 #include "../include/czmq.h"
 
-enum proxy_socket
+typedef enum proxy_socket
 {
     NONE = -1,
     FRONTEND,
     BACKEND,
     SOCKETS
-};
+}
+proxy_socket;
 
 #define AUTH_NONE  0
 #define AUTH_PLAIN 1
--- a/src/zproxy.c
+++ b/src/zproxy.c
@@ -24,6 +24,18 @@
 
 #include "../include/czmq.h"
 
+enum proxy_socket
+{
+    NONE = -1,
+    FRONTEND,
+    BACKEND,
+    SOCKETS
+};
+
+#define AUTH_NONE  0
+#define AUTH_PLAIN 1
+#define AUTH_CURVE 2
+
 //  --------------------------------------------------------------------------
 //  The self_t structure holds the state for one actor instance
 
@@ -33,6 +45,10 @@ typedef struct {
     zsock_t *frontend;          //  Frontend socket
     zsock_t *backend;           //  Backend socket
     zsock_t *capture;           //  Capture socket
+    int auth_type[SOCKETS];     //  Auth type for sockets
+    char *domain[SOCKETS];      //  Auth domains for sockets
+    char *public_key[SOCKETS];  //  Public keys for sockets
+    char *secret_key[SOCKETS];  //  Secret keys for sockets
     bool terminated;            //  Did caller ask us to quit?
     bool verbose;               //  Verbose logging enabled?
 } self_t;
@@ -46,6 +62,14 @@ s_self_destroy (self_t **self_p)
         zsock_destroy (&self->frontend);
         zsock_destroy (&self->backend);
         zsock_destroy (&self->capture);
+
+        for (int i = 0; i < SOCKETS; i++)
+        {
+            zstr_free (&self->domain[i]);
+            zstr_free (&self->public_key[i]);
+            zstr_free (&self->secret_key[i]);
+        }
+
         zpoller_destroy (&self->poller);
         free (self);
         *self_p = NULL;
@@ -65,9 +89,8 @@ s_self_new (zsock_t *pipe)
     return self;
 }
 
-
 static zsock_t *
-s_create_socket (char *type_name, char *endpoints)
+s_create_socket (self_t *self, char *type_name, char *endpoints, proxy_socket selected_socket)
 {
     //  This array matches ZMQ_XXX type definitions
     assert (ZMQ_PAIR == 0);
@@ -85,6 +108,28 @@ s_create_socket (char *type_name, char *endpoints)
     }
     zsock_t *sock = zsock_new (index);
     if (sock) {
+        if (self->domain[selected_socket]) {
+            // Apply authentication domain
+            zsock_set_zap_domain (sock, self->domain[selected_socket]);
+        }
+        if (self->auth_type[selected_socket] == AUTH_PLAIN) {
+            // Enable plain authentication
+            zsock_set_plain_server (sock, 1);
+        }
+        else
+        if (self->auth_type[selected_socket] == AUTH_CURVE) {
+            // Apply certificate keys
+            char *public_key = self->public_key[selected_socket];
+            assert(public_key);
+            char *secret_key = self->secret_key[selected_socket];
+            assert(secret_key);
+            zsock_set_curve_publickey (sock, public_key);
+            zsock_set_curve_secretkey (sock, secret_key);
+
+            // Enable curve authentication
+            zsock_set_curve_server (sock, 1);
+        }
+
         if (zsock_attach (sock, endpoints, true)) {
             zsys_error ("zproxy: invalid endpoints '%s'", endpoints);
             zsock_destroy (&sock);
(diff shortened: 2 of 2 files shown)


jemc:
Note the CI errors about explicitly specifying `enum` (or making a `typedef`) and about the implicit declarations (you need to define your functions in a different order or use forward declarations).

ritchiecarroll:
Sorry, I made the changes in Visual Studio where they compiled OK - didn't test any other platforms. Also, very new to GitHub - so, should this pull request be rejected where I can make changes and try again?

jemc:
@ritchiecarroll - no worries.  We don't need to close or reject this pull request.  We're actually quite permissive about pull requests in this project (see [the C4 RFC](http://rfc.zeromq.org/spec:22)) but we do require (among other things) that the code compiles and passes the test suite on the major supported platforms.

To add fixes for the noted compiler errors, you can simply push more `git` commits to your `master` branch.  This pull request ticket will automatically update to reflect the new commits after you push them to your fork on GitHub (just as you pushed this commit).

Let us know if you have any more questions, and thanks for contributing!

jemc:
Also, in case it wasn't obvious, when I talk about CI errors, I mean Travis CI, our continuous integration testing server.  You can see the build info by clicking on the `X` failure icon.  From there you can navigate to see console output from each job item in the build matrix, for example, this output from the most recent commit: https://travis-ci.org/zeromq/czmq/jobs/80377090

hintjens:
@ritchiecarroll awesome!

Let me quickly merge this...

hintjens:
... ok, there are a bunch of things I'd like to clean up in the code. Variables like 'i', spacing and brackets, etc. If you read the CLASS style guide or existing code you'll see the differences. You can take a shot at it or I'll do it later.

ritchiecarroll:
@hintjens Would like to try the clean up, will read the CLASS style guide. Do you think adding security to the CAPTURE socket would also be useful? I had considered it while making updates but was afraid the test code was getting onerous - even so, adding this command option would complete the security implementation of zproxy.
@jemc  Since Travis is very unhappy with me and after the merge the code I added has broken the build, do I just create a new pull request to apply fixes? Also, when I use cmake on Linux I don't see the same level of issues as the CI engine - is there a local build script I should run to be equally as picky as Travis?

hintjens:
No, don't solve problems you haven't hit yet...

jemc:
But *do* solve the compiler problems that broke the build :wink:.

hintjens:
Yes, indeed.

ritchiecarroll:
OK - added another pull request that should meet style guide. Runs on Windows and Linux, however, Travis may find more items, guess we'll see: https://github.com/zeromq/czmq/pull/1091

ritchiecarroll:
@jemc Looks like some else also edited the same file, zproxy.c, before I did a pull request - there is a message to the effect of "branch has conflicts that must be resolved".  I the guy before me just make a quick change to get it to compile in his environment. I had also already made a similar change. Is it on me to figure out how to fix this - or do you handle the merge upstream?

hintjens:
You pull the upstream master back to your fork clone, fix any conflicts,
and push again. If needed, do a push --force to your fork repo.

jemc:
@ritchiecarroll - there are various ways to do this - search for something like "resolving merge conflicts with git" on the web and you will see various tutorials.

ritchiecarroll:
Thank you for all the assistance. FYI, pull request should be ready to merge: https://github.com/zeromq/czmq/pull/1091 - note that Travis CI thinks there is an issue with ZMQ_REPO=zeromq4-x WITH_LIBSODIUM=1: https://travis-ci.org/zeromq/czmq/builds/80500309 - all other builds seem to be working OK. This issue occurs during the zproxy authentication testing, but the issue looks to be a little deeper than the simple zproxy actor code and correction is likely going to be outside of my contribution availability.

ritchiecarroll:
I keep poking at this issue - I cannot reproduce locally, however, Travis CI for certain build configurations is fairly consistent about having a condition related to a PUSH socket and auth blacklisting - seems to just hang, likely waiting for something, race condition?

I was wondering is a PUSH socket configured with DENY auth is completely backwards from a use case perspective and should not be tested anyway? Does this ring a bell as a logic error in the test which would explain the occasional trouble? Could exclude DENY test for BACKEND socket and just test ALLOW/DENY with FRONTEND PULL socket.

I am happy to keep trying things, but the only way I have to test is to abuse Travis. I feel like I am generating too much noise with the pull requests...

hintjens:
Feel free to abuse Travis by pushing commits to your fork. I know no other
way to test this either, adding traces etc. You can push temporary commits,
and wipe them with --force. Or you can push to a branch if you like that
kind of stuff.

It could well be a race condition. As for DENY on the backend... not sure.

ritchiecarroll:
Thanks - just to verify, are suggesting I attempt to setup Travis on my fork or just keep a pull request open on czmq?

jemc:
@ritchiecarroll - either is okay.  If you do it with an ongoing pull request, just be sure to mark it as a work in progress in a comment so that others don't merge it prematurely.

Asmod4n:
Travis is great when it comes to race conditions, because their free boxes are always overbooked :)

ritchiecarroll:
@Asmod4n - yeah, I think that may be the key issue as to why I cannot reproduce locally.

@hintjens  - After getting Travis CI working against my local fork and doing some testing, here is my working theory for the issue: the zproxy API doesn't currently expose a way to get the bound port for its sockets, as a result the test code must find an available port then use this for the frontend and backend sockets as well as connecting clients. The port can be found either by random selection or through ephemeral binding port (tried both). In either case the port gets bound to a socket, then immediately unbound so that the proxy can use it. Although there is a possible minor race condition with another concurrent CI build grabbing the port during this transition, the bigger issue seems to be the loading on the CI servers. I don't think the system has always completely released the port during the transition meaning that sometimes the the binding will fail in the proxy socket configuration even though the binding just succeeded while finding an available port. Here are some thoughts on resolution:

* One option would be to add a command to "return" the bound port from the API and use ephemeral binding for test code like the zauth actor does - this all assuming the actor APIs can support return data, which seems like it would work OK. However, this option is complicated by the fact that that the zproxy code uses zsock_attach which can attach to multiple endpoints and the zock_attach function does not currently provide a way to return the bound ports - but looks like it could be modified to return an array of bound port numbers as an out parameter - a breaking change unless this was internal only code or a new or overloaded function.
* Another option might be to inject a delay between the port transitions - this is very non-deterministic as far as timing is concerned and increases the chances of an another CI build grabbing the port in the mean-time, however, it may have a chance at working with the existing code. It would also have the unfortunate side effect of increasing build time with injected delays.
* Another not as comprehensive testing option might be to skip the ALLOW/DENY testing which need an IP and stick with inproc sockets for all remaining security testing - assuming PLAIN and CURVE work OK on inproc sockets.
* A very bad option would be to make the tests use fixed ports and limit CI concurrency to one.

ritchiecarroll:
Addendum - a quick test of inject a one second delay between port bing transitions seems very effective at resolving the errors...

jemc:
@ritchiecarroll - can you not use `inproc` to avoid this issue?

jemc:
Ah, I see you noted that in your list already.
> Another not as comprehensive testing option might be to skip the ALLOW/DENY testing which need an IP and stick with inproc sockets for all remaining security testing - assuming PLAIN and CURVE work OK on inproc sockets.

ritchiecarroll:
UPDATE: (Note there are questions at the bottom of this message) Having seemed to corrected the CI testing "errors", I circled back to the issue where the CI testing seems to occasionally lockup during a CI run in the security expanded zproxy testing. The issue is in the infinite wait in the zproxy constructor (-1 value in zpoller_wait call):
```
void
zproxy (zsock_t *pipe, void *unused)
{
    self_t *self = s_self_new (pipe);
    assert (self);
    //  Signal successful initialization
    zsock_signal (pipe, 0);

    while (!self->terminated) {
        zsock_t *which = (zsock_t *) zpoller_wait (self->poller, -1);
        if (zpoller_terminated (self->poller))
            break;          //  Interrupted
        else
        if (which == self->pipe)
            s_self_handle_pipe (self);
        else
        if (which == self->frontend)
            s_self_switch (self, self->frontend, self->backend);
        else
        if (which == self->backend)
            s_self_switch (self, self->backend, self->frontend);
    }
    s_self_destroy (&self);
}
```
I set the poller wait call to timeout after 1/2 second and everything now works fine (https://travis-ci.org/ritchiecarroll/czmq/builds/80879574), here's the fix:
```
void
zproxy (zsock_t *pipe, void *unused)
{
    self_t *self = s_self_new (pipe);
    assert (self);
    //  Signal successful initialization
    zsock_signal (pipe, 0);

    while (!self->terminated) {
        zsock_t *which = (zsock_t *) zpoller_wait (self->poller, 500);
        if (zpoller_expired (self->poller))
            continue;       // Timed out
        if (zpoller_terminated (self->poller))
            break;          //  Interrupted
        if (which == self->pipe)
            s_self_handle_pipe (self);
        else
        if (which == self->frontend)
            s_self_switch (self, self->frontend, self->backend);
        else
        if (which == self->backend)
            s_self_switch (self, self->backend, self->frontend);
    }
    s_self_destroy (&self);
}
```
Question 1: Is there any reason why this function should really wait infinitely? 

Question 2: If not, is 500ms a good wait timeout, or is something smaller better?

Question 3: In the z_proxy self_t opaque structure, the terminated flag is used by more than one thread (this could have been the real culprit), does ZMQ have a threading policy for not marking fields as volatile?

Thanks,
Ritchie

jemc:
> In the z_proxy self_t opaque structure, the terminated flag is used by more than one thread

Can you explain this?  This should not be happening at all.

ritchiecarroll:
@jemc Sorry, I think you are right. I had assumed incoming commands may be on another thread than the zactor loop thread since calls were coming from testing thread - but after a more careful look, I think that's all on the same thread. Please ignore Question 3.

jemc:
Well, it seems to me that waiting for messages with no timeout is the most optimal way to wait, as otherwise you waste CPU time and context switching waking up threads that have no reason to wake up.

I don't have time right now to look at it in detail, but it seems like if this pattern of waiting is causing problems, you have some other race condition in the code or the tests, and while adding a timeout is an effective workaround, it's not a solution to the underlying problem.

ritchiecarroll:
Agreed. At the moment I can only test via Travis CI (no local reproducible issue), which makes debugging a very long process. I'll change the timeout to one second to minimally impact CPU waste and do a PR since this will make sure CI is working fine. As I have time I will try to dig deeper into the race condition, it may well go beyond the zproxy class.

ritchiecarroll:
Reverted back to infinite timeout since after more CI testing even with a timeout, the system will still sometimes hang. The only sure workaround for a CI build is to only test the FRONTEND socket. The current PR comments out BACKEND socket testing to allow for clean builds. I will create an issue for resolution of testing timeout and provide known details.

ritchiecarroll:
Interesting new clue: same behavior occurred during CI build for last merge, however, the issue this time is with zuth, not with zproxy: https://travis-ci.org/zeromq/czmq/jobs/80900844

The zproxy test code is a step-by-step match of zuth but executed over three modes, i.e., frontend, backend and frontend+backend.

Issue may lie with authentication testing pattern and/or a lower level issue.

PR #1088, opened 2015-09-15
Minor fixes

sappo:


jemc:
The build is broken, but it's not this PR's fault :wink:.

sappo:
Hmm...weird, why isn't that error showing on my machine?

jemc:
You probably need the compiler flag `-Werror=format-security`.

sappo:
Okay, is this a default in the travis containers.

jemc:
As far as I know, it is a default in Ubuntu.

sappo:
What do you think about explicitly settings this flag?

PR #1089, opened 2015-09-15
Problem: The CFLAG '-Werror=format-security' is used by default in some

sappo:
systems in particular the CI-Server. Which results in test build failures if this flag is not set on the development system.
Solution: Activate the flag by default to find these compiler errors early.

Also: fixed the failing zproxy build!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,6 +5,9 @@
 
 ACLOCAL_AMFLAGS = -I config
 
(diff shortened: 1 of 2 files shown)


jemc:
Won't the `Makefile.am` change be clobbered by zproject, though?

sappo:
Yes :-D
I'll add it there too.

PR #1091, opened 2015-09-15
Updated code to better meet style guide.

ritchiecarroll:
Code clean up to better match CLASS style guide.


jemc:
I'm going to go ahead and merge, though I think the problem seen on the two failed matrix items may be due to issues with the `zproxy` test here.

Likely it has some sort of race condition or other non-determinism that causes it to succeed most of the time, but fail sometimes.  It probably needs more investigation.

PR #1092, opened 2015-09-16
Updated ztrie.c with a strndup function so it will compile on Windows

ritchiecarroll:
The file "ztrie.c" does not compile on Windows due to lack of a "strndup" function. Added one to source file - there's likely a better location for the function, but this gets things working.

### DIFF
--- a/src/ztrie.c
+++ b/src/ztrie.c
@@ -39,6 +39,25 @@
 #define MIN_LEN(x,y) \
     y + ((x - y) & ((x - y) >>(sizeof(int) * CHAR_BIT - 1)))
 
+// TODO: Move to a more appropriate location:
+#if defined (__WINDOWS__)
+char *
+strndup (const char *s, size_t size) {
+    char *dup;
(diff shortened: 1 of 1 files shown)


sappo:
We do have a zstr class which could be extended with cross platform string function. Thanks for the pull request though, I don't have a windows machine to test on.

hintjens:
We've avoided the use of strndup elsewhere (and I'd avoid this, it's not
portable). If you really want to define non-portable functions, the two
places we do that (at present) are either as macros in czmq_prelude.h or as
methods in zsys.

sappo:
I'll replace the strndup function. I didn't know it is not portable.

sappo:
What would be a suitable substitution that works cross platform?

hintjens:
Usually just strdup, using pointers rather than fixed size strings.

sappo:
Hmm...that's not really an option as I copy tokens from a larger url/uri/... .

PR #1093, opened 2015-09-16
Problem: strndup is not cross platform

sappo:
Solution: use own implementation of strndup which works cross platform.

@ritchiecarroll could you please verify if this works for windows as well. Thanks!


ritchiecarroll:
Yes - compiles on my Windows environment...

PR #1100, opened 2015-09-17
If interrupted in a middle of a multipart send or receive sequence, keep

maxkozlovsky:
retrying until complete message is send/received.
Fix the frame leak if a message send fails, do not remove a frame from the
message list until the send has completed successfully.


jemc:
The CI failures are unrelated to this PR - I'm merging.

PR #1106, opened 2015-09-19
Problem: zframe_strhex is needlessly using zmalloc

Asmod4n:
Solution: change it to malloc.


Asmod4n:
Error is unrelated.

sappo:
@Asmod4n What's the difference anyhow? Is there a performance loss?

Asmod4n:
@sappo zmalloc nullifies the buffer, which has a performance hit. But since zframe_strhex doesn't read the buffer there is no need to do that.

sappo:
@Asmod4n thanks :+1:

PR #1110, opened 2015-09-26
Problem: zproject changed the known dependency name for zmq to libzmq

sappo:
Solution: change the dependency name and regenerate zproject files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ endif()
 list(APPEND CMAKE_MODULE_PATH "${SOURCE_DIR}")
 
 ########################################################################
-# ZMQ dependency
+# LIBZMQ dependency
 ########################################################################
 find_package(ZeroMQ REQUIRED)
 IF (ZEROMQ_FOUND)
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,12 +9,12 @@ AM_CFLAGS = \
 	-Werror=format-security
 
 AM_CPPFLAGS = \
-    ${zmq_CFLAGS} \
+    ${libzmq_CFLAGS} \
     ${uuid_CFLAGS} \
     -I$(srcdir)/include
 
 project_libs = \
-    ${zmq_LIBS} \
+    ${libzmq_LIBS} \
     ${uuid_LIBS}
 
 SUBDIRS =
--- a/builds/qt-android/build.sh
+++ b/builds/qt-android/build.sh
@@ -30,20 +30,20 @@ if [[ $ANDROID_BUILD_CLEAN ]]; then
 fi
 
 ##
-# Make sure zmq is built and copy the prefix
+# Make sure libzmq is built and copy the prefix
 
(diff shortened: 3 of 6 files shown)


bluca:
Thanks to this fix now the pkg-config for uuid finally works, which is nice! Downside is that Android build fails because on that system uuid.h is not in a uuid subdirectory:

$ find android-ndk-r10e/ -name "uuid.h"
android-ndk-r10e/platforms/android-21/arch-arm/usr/include/linux/uuid.h
android-ndk-r10e/platforms/android-21/arch-arm64/usr/include/linux/uuid.h
android-ndk-r10e/platforms/android-21/arch-x86_64/usr/include/linux/uuid.h
android-ndk-r10e/platforms/android-21/arch-mips64/usr/include/linux/uuid.h
android-ndk-r10e/platforms/android-21/arch-mips/usr/include/linux/uuid.h
android-ndk-r10e/platforms/android-21/arch-x86/usr/include/linux/uuid.h

It didn't use to fail before because pkg-config didn't work at all, and the fallback code in configure.ac failed as well since on Android there is no SO to link against, just a header, so it always built without uuid.h.

I'm having a CI run on a fix, will do a PR later if it works. It will have to be fixed with an IFDEF in the public header czm_prelude.h unfortunately (there are already a couple for uuid.h).

bluca:
Broader question maybe is whether the include should be plain and simple <uuid.h> everywhere, and use the CPPFLAGS to pass the subdirectory. pkg-config is already used for that, and we could change the configure.ac to define a similar flag and pass the subdir in this snippet:

```
        if test "x$search_libuuid" = "xyes"; then
            if test -r "${with_libuuid}/include/uuid/uuid.h"; then
-                uuid_synthetic_cflags="-I${with_libuuid}/include"
+               uuid_synthetic_cflags="-I${with_libuuid}/include/uuid"
                uuid_synthetic_libs="-L${with_libuuid}/lib -luuid"
            else
```

But since that changes how the existing public header works, that would probably break the ABI I guess?

jemc:
@bluca - not sure I followed all your suggestions, but thanks for keeping an eye on the qt-android build.  I'm okay with adding special case code there to treat uuid however it needs to be treated on android.

bluca:
@jemc - sorry I was in a hurry when I typed :-)

I opened https://github.com/zeromq/zproject/pull/246 to address the immediate problem of the build, will regenerate and do a PR against CZMQ once the maintainers are happy with that.

In general, as I pointed out here: https://github.com/zeromq/zproject/pull/232 the current zproject autoconf generation assumes that a project must link against all dependencies. But this assumption does not work with the Android NDK, as many libraries (including uuid) are included in the C library, and only a (compatible) header is exposed. So the autoconf test tries to link when looking for the library and fails (and even if it succeeded, it unconditionally adds -l<LIBNAME> to LDFLAGS, which would cause a build failure with the NDK).

So there's more work to do there. I'll try to get around and play with it when I have time, but no promises :-)

PR #1111, opened 2015-09-27
Regen from zproject to fix qt-android

bluca:
Recent commit fixed the broken uuid pkg-config check, but this in turn exposed another bug in the qt-android build system.


bluca:
This is just a regeneration from latest Zproject, so I assume it's ok to commit the changes to src/Makemodule.am as well.

bluca:
Build failures are unrelated, usual flaky tests on the overloaded CI machines :-)

Issue #1094, opened 2015-09-16
Problem: czmq doesn't have any CI test for windows


sappo:
Hence Travis CI doesn't support windows we cannot test directly on a windows host, but we can cross compile the code.

@pijyoi Do you still have a running build using MXE to cross compile for mingw. All my efforts failed so far :(

pijyoi:
I just tried it out with a fresh build of MXE. mingw32 got dropped, have to
use MinGW-w64.

compile zeromq-4.1.2
====================
./configure --host=i686-w64-mingw32.static --prefix=$HOME/mingw32
--without-libsodium
make
make install

compile czmq master
===================
export PKG_CONFIG_PATH_i686_w64_mingw32_static=$HOME/mingw32/lib/pkgconfig
CFLAGS="-DLIBCZMQ_EXPORTS -D__USE_MINGW_ANSI_STDIO" ./configure
--host=i686-w64-mingw32.static --prefix=$HOME/mingw32
sed -i.bak s/-pedantic// Makefile
make
make install

notes
=====
make clean fails
make[1]: zproject_mkman: Command not found


# removal of -pedantic to fix the following
src/zsock.c: In function 'zsock_vsend':
src/zsock.c:677:32: error: ISO C does not support the 'I64' ms_printf
length modifier [-Werror=format=]
             zmsg_addstrf (msg, "%" PRIu64, va_arg (argptr, uint64_t));

# define __USE_MINGW_ANSI_STDIO to fix the following
trie.c: In function 's_ztrie_print_tree_line':
src/ztrie.c:614:21: error: unknown conversion type character 'z' in format
[-Werror=format=]
             printf ("%s (params: %zu, endpoint: %s, type: %d)\n",
self->token,
                     ^
src/ztrie.c:614:21: error: format '%s' expects argument of type 'char *',
but argument 3 has type 'size_t {aka unsigned int}' [-Werror=format=]
src/ztrie.c:614:21: error: format '%d' expects argument of type 'int', but
argument 4 has type 'char *' [-Werror=format=]
src/ztrie.c:614:21: error: too many arguments for format
[-Werror=format-extra-args]

sappo:
Thanks, do you know which packages you'd installed for MXE. A complete build would probably unsuitable for a test run ;)

zproject_mkman is installed by zproject so the failing clean is not that bad.

The new "-Werror=format=" causes the ISO C errors as window/mingw is not ISO C, so it's okay to disable it.

[...]

PR #1125, opened 2015-10-17
Add Windows Visual Studio CI appveyoyer.yml file

madebr:
In #1094, a CI test for Windows was requested.
Appveyor provides a framework for Visual Studio on Windows.
See http://www.appveyor.com/

Notes:
- Only tests Visual Studio 2015
- It builds 32- and 64-bit
- It builds Release and Debug
- It will build the libzmq dependency everytime
- It does not build and run the libzmq tests
- It will run the selftest of czmq after the build
- It will produce a zip with the built files, which can be downloaded
  from the artifacts page.
- Remote debugging (using rdesktop) is possible. See
  http://www.appveyor.com/docs/how-to/rdp-to-build-worker
  on how to pause the build and rdp to the worker.

Known problems:
- In FindUuid.cmake, a bad path is set. So UUID is not used in Visual Studio.
- No output of the czmq_selftest
- The self test fails at an assertion. For an unknown reason, the Release fails
  and Debug succeeds.
- The self test fails. An assertion can cause a dialog to appear, which can time out the test.

Sample output:
- https://ci.appveyor.com/project/madebr/czmq/build/git-980628ee
  (the self test times out)
- https://ci.appveyor.com/project/madebr/czmq/build/build-108
  (I've re-run the same test, that's why an error appears at the top because no 2 builds can have the same names)

Please reply when you would have some remarks.

### DIFF
--- a/FindUuid.cmake
+++ b/FindUuid.cmake
@@ -3,11 +3,16 @@
 #  Please refer to the README for information about making permanent changes.  #
 ################################################################################
 
+if (NOT MSVC)
 include(FindPkgConfig)
 PKG_CHECK_MODULES(PC_UUID "uuid")
 IF (NOT PC_UUID_FOUND)
     PKG_CHECK_MODULES(PC_UUID "uuid")
 ENDIF (NOT PC_UUID_FOUND)
+else (NOT MSVC)
+set(PC_UUID_INCLUDE_DIRS "C:/Program Files")
+set(PC_UUID_LIBRARY_DIRS "C:/Program Files")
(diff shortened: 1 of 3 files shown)


sappo:
@madebr this is great. 

Can we integrate this like Travis CI?

sappo:
I managed to add a web hook to Appveyor and inserted a build status to the readme.

sappo:
... and its me again ;)
@madebr the development configuration works fine but the release builds fail even though the tests pass successful.

Issue #527, opened 2014-06-17
zdir_test sensitive to current working directory


scpeters:
The czmq selftest was failing on mac osx in a [homebrew pull request](https://github.com/Homebrew/homebrew/pull/30243) to upgrade to 2.2.0. In testing on my local machine, I found that `zdir_test` is sensitive to the current working directory of the test process. I managed to fix the test in https://github.com/scpeters/homebrew/commit/f1cdcfebd5e7f699db5d1ee3e2eb9d2186620f8f by changing the current working directory before launching the selftest.

I don't know if the test needs to change, but maybe the requirements for the current working directory for selftest could be documented somewhere.

bluca:
Closing as it was fixed by https://github.com/zeromq/czmq/pull/1127

PR #1127, opened 2015-10-19
fix zdir traversing too high

samrussell:
Bug https://github.com/zeromq/czmq/issues/527

czmq_selftest calls zdir_test, which enumerates the current and parent working directories. This breaks `make check` when the parent directory contains files with chmod 600 owned by other users (problematic if the build directory is somewhere shared, like /tmp/)

Since czmq_selftest is called in the czmq base directory by `make check`, it makes sense for this test to refer to "src" and "." instead of "." and "..", preventing zmq from playing outside its build directory

### DIFF
--- a/src/zdir.c
+++ b/src/zdir.c
@@ -936,13 +936,13 @@ zdir_test (bool verbose)
     printf (" * zdir: ");
 
     //  @selftest
-    zdir_t *older = zdir_new (".", NULL);
+    zdir_t *older = zdir_new ("src", NULL);
     assert (older);
     if (verbose) {
         printf ("\n");
         zdir_dump (older, 0);
     }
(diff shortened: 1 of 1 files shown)


sappo:
Thanks for the fix!

Would you please read the [contributors guideline] (https://github.com/zeromq/czmq/blob/master/CONTRIBUTING.md) and adjust your commit messages with `git commit --amend` accordingly. :+1:

samrussell:
Hi Sappo,

The contributors guidelines are quite verbose, could you please suggest an alternative commit message?

Cheers
Sam

hintjens:
@samrussell simply add "Problem:" at the start of the commit message. You did it right otherwise.

samrussell:
@hellermf I've created pull request #1130 to address your problem, please let me know if this helps

PR #1128, opened 2015-10-19
Use timeout on Windows CI for czmq_selftest

madebr:
This fixes my last Windows CI PR.
False detection of a non-finishing test will not cause a failed test.

Anyhow, czmq_selftest has some issues.
- On Windows, not all output is redirected into the web log (no subtest is shown at all)
- Assertion errors on windows are not shown in the web log
- On Linux, redirection of the output of czmq_test will also nog show all output. (#1129)

On my system, the second one will not show all output. Maybe the issues are related.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@ os: Visual Studio 2015 #  Windows Server 2012
 
 environment:
   CMAKE_GENERATOR: "Visual Studio 14 2015"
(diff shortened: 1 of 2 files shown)


sappo:
@madebr maybe we should rename timeout.py to appveyor_testrunner.py to clarify its purpose.

madebr:
You can do that.
It also works on my Linux machine.

PR #1131, opened 2015-10-22
Use CTest for tests and added functionality to czmq_selftest

madebr:
- Make the function signatures of every test function the same: `void test_XXX(bool)`
- Give functionality to `czmq_selftest`:
  * `-v`: verbose output
  * `-e`: for windows, don't display a error dialog when an assertion is hit
  * `--nb`: print how many tests are available
  * `--list`: print all available tests to stdout
  * `--test`: run 1 particular test
- Use CTest as test framework for appveyor
  * This eliminates the python test
  * Time limits are configured in CMakeLists.txt
  * Every test is run seperately (in addittion to all of them as a whole)
  * CTest knows about the name of the test by parsing `czmq_selftest.c`.
    It's important to have `DECLARE_TEST` wrapped around the function name.
  * Verbose output of all tests

In previous issues, I mentioned already that `czmq_selftest` has some issues on Windows.
By failing the build explicitly if a subtest fails, I hope to attract some attention to them.
(Previously, the return code of a process could be ignored)

A problem I struggle with, is that the output of the test is empty when an assertion hits.

Therefore, I have created an issue to enumerate the build errors: see #1132 .


sappo:
I like the changes to the selftests. What are the dependencies to CMAKE though? IMO there should not be any dependencies to only one build system.

One more remark: If you read the contributors agreements you'll notice that each commit should give a statement about the Problem and the Solution. You can have a look at the commit history to get an example. It would be nice if you can apply this format to future commits. Thanks!

madebr:
- The dependency lies in the way the tests should be declared inside `src/czmq_selftest.c`.
  If the format does not match, the test will not be discovered by CTest.
  The current naming of things is maybe not so good.
  I'll change it to something general where the name CMAKE is not present.
- Can you give a suggestion about a good commit message?

bluca:
Hello,

Will the changes to the signatures marked with CZMQ_EXPORT warrant a bump of the SO version?

madebr:
I probably should, I did not think about that.

madebr:
I addressed the build system dependency issue of @sappo.
The tests need to be wrapped within DECLARE_TEST in src/czmq_selftest.c

sappo:
@madebr Regarding the commit messages, e.g. for 0a25844 I would phrase it:
```
Problem: It's not possible to run a class selftest independently of the other class selftests.
Solution: Add a new option to the arguments of czmq_selftest to select a specific class selftest to run.
```

sappo:
There is still one issue to address. The czmq_selftests are generated by zproject. Therefore we would need to apply these change there as well. My proposal is to test these changes in czmq and if we are satisfied with them apply them to zproject. Any objections?

madebr:
i would love to modify the czmq_selftest generator, but do not find the generator.
Can you generate a link for me?

madebr:
I made the changes to the commit messages as requested.

- Additionally, I added the sodium dependency on Windows CI.

hintjens:
@madebr ah, I was looking for your github user id. You're using anonymous.maarten@gmail.com in some commits, which GitHub does not recognize.

hintjens:
@madebr could you please read http://rfc.zeromq.org/spec:21 and use this style for your patches? Thanks!

PR #1135, opened 2015-10-26
Sync czmq with zproject

madebr:
This PR will apply the fixes of my latest PR to zproject to czmq.

problem: cmake does not work due to being out-of-sync with zproject
solution: run gsl and sync it


hintjens:
Thanks for the stream of useful patches :-)

PR #1137, opened 2015-10-27
Problem: several sources break style with CLASS

hintjens:
Specifically:

* they use inline declarations in for loops, which is not standard C
* they use the variable name 'i' in loops, which is banned

I've no idea how the inline declarations are passing. Perhaps someone
has tweaked zproject to allow this on gcc. On MSVC since we use C++
it will not be caught.

Solution: fix the sources.

### DIFF
--- a/src/zarmour.c
+++ b/src/zarmour.c
@@ -681,6 +681,7 @@ s_armour_test_long (zarmour_t *self, byte *test_data, size_t length, bool verbos
 {
     if (verbose)
         zarmour_print (self);
+
     char *test_string = zarmour_encode (self, test_data, length);
     assert (test_string);
     if (verbose)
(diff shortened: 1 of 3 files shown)


pijyoi:
Inline declarations in for loops is C99, isn't it?

Asmod4n:
@pijyoi yes, the outside declaration is c89.

hintjens:
Indeed...  I need to find the original reason we banned this. We do target
C99 after all.

sappo:
Maybe you banned this because you can change the scope for inline declarations for c++ with the compiler flags `-ffor-scope` and `-fno-for-scope`?

hintjens:
I don't play with such options...  It seems to be my error.  I'll modify
CLASS and fix the code.

PR #1143, opened 2015-10-30
Problem: zframe fails to compile, missing ifdef

bluca:
Solution: add ifdef ZMQ_SERVER around calls to zmq_msg_set_routing_id,
which is only available in libzmq 4.2 and later


hintjens:
Thanks for the fix. I was getting around to it and got distracted by other things... :)

bluca:
No problem! One more for libzmq 2 build in the pipeline :-)

PR #1146, opened 2015-10-30
Fix check-py failure

bluca:
@hintjens - PR #1139 removed the implementation of zframe_send_reply, but it was left in the api XML and so in the generated bindings and headers. This causes check-py to fail.

This PR removes the definition from the XML of zframe and regenerates the bindings accordingly. Is this the right thing to do, or is the implementation just out temporarily? Please decline if that's the case :-)

I've also updated the documentation. Are the .doc files somehow autogenerated or should they be updated manually?


hintjens:
Sorry for my half-assed changes. Yes, removing the method from the XML API is right.

The doc files are built by "make all" in the doc subdirectory.

PR #1151, opened 2015-11-05
Problem: Unable to compile against latest libzmq.

twhittock:
Solution: Use correct function name when compiling ZMQ_SERVER related code


twhittock:
not sure what happened there - i must have been using some odd version of libzmq. ignore, sorry.

Issue #1152, opened 2015-11-12
Problem: no Java native interface exists to CZMQ


trevorbernard:

hintjens:
Done in zproject.

PR #1153, opened 2015-11-13
Issue #1152 - First working version of CZMQ JNI binding

trevorbernard:


jemc:
I have my doubts about including bindings that are not code-generated into the main CZMQ repository - I worry mainly that they cannot be kept up to date in the future and we'll have dead code lying around.  Also, having some bindings be generated and others not generated is likely to cause confusion among contributors.

If these bindings can't be generated via zproject, why not create a separate repository for these, as is already common for handwritten zeromq-related bindings to other languages?

jemc:
By project guidelines, I will merge anyway, but I have serious doubts about this PR.

trevorbernard:
It was per @hintjens request. The goal is to create a working version then slowly replace with code generation

jemc:
:+1:

hintjens:
Yes, this is designed obsolescence...  :)

PR #1165, opened 2015-11-19
problem: there was no gradle task to publish the czmq-jni jar locally

awynne:
solution: added maven-publish plugin and configuration to enable a the standard publishToMavenLocal task


sappo:
@awynne this file is generated by zproject zproject_bindings_jni.gsl. If you like this change to be persistent you would need to add it there. Thereby other projects like zyre can leverage from your patch automatically.

sappo:
@awynne just FYI you can avoid those merge commits if you do a git pull --rebase

awynne:
Got it.  Thanks for the tips!

Issue #1164, opened 2015-11-19
zsock_option functions as stand-alone API class or embedded within zsock class?


paddor:
I'm improving the generated Ruby bindings of CZMQ. I've noticed that the zsock_option functions ([documentation](http://api.zeromq.org/czmq3-0:zsock-option)) don't have a corresponding XML API file. So I created a GSL file which does that, as the class `zsock_option` (at least for the major ZMQ version 4). See [comparison](https://github.com/zeromq/czmq/compare/master...paddor:better_ruby_bindings?expand=1).

However, I've hit a problem. Those functions aren't prefixed with the expected class name `zsock_option_`, but `zsock_` instead.

I hesitated adding these functions to the existing `zsock` class XML API, because those functions are documented to
> [...] take a void * "polymorphic" reference, which can be either a zsock_t or a zactor_t reference, or a libzmq void *.

I could just add them to the existing `zsock` class XML API. To set any options on a zactor_t object using a binding later, one would have to create an instance of the Zsock class with a given zactor_t pointer or something.

I thought it'd nicer to have these functions grouped in a ZsockOption class which can be used for any of the three supported reference types, either by exposing them all as singleton methods. To do this, the `<class>` tag in the XML API files would have to support a new attribute, like `function_prefix = "zsock_"`. A nicer approach would be generating a Ruby module which can be mixed into the Zsock and Zactor classes (although this would require additional information from the API file).

Maybe the Zactor functionality doesn't make a lot of sense in a high-level binding anyway, so one could ignore that requirement, which makes the integration directly into the `zsock` class sound even better.

Which approach sounds better to you? Any other ideas?

paddor:
Huh, I've just seen that some of the instance methods defined in the API for zsock have an attribute `polymorphic = "1"`. I guess this means that the polymorphic zsock_option functions can and should be added there.

jemc:
The `polymorphic` option was recently added, and as far as I know, only implemented in the python bindings so far, as it was the motivation to support it was from users/authors of those bindings.  The ruby bindings (and others) would have to be updated to support `polymorphic`.

In python, this is easier, since all instance methods can also be called as singleton/class methods with an explicit `self` object passed in as the receiver instance.

My first thought about the best way to support `polymorphic` in ruby is to imitate the python behavior:
1. Define a singleton method where the first argument is the polymorphic receiver object (the explicit "self") - `def self.function_name(receiver, arg1, arg2)`.
[...]

PR #1168, opened 2015-11-20
Include zsock_option functions into zsock.xml

paddor:
This also fixes paths to the license file and actually turns on the `<include>` tag within class API XML files.

This is related to #1164.

### DIFF
--- a/api/zactor.xml
+++ b/api/zactor.xml
@@ -1,7 +1,7 @@
 <class name = "zactor">
     actor
 
-    <include filename = "../license.xml" />
+    <include filename = "license.xml" />
 
     <callback_type name = "fn">
         Actors get a pipe and arguments from caller
--- a/api/zdir.xml
+++ b/api/zdir.xml
@@ -1,7 +1,7 @@
 <class name = "zdir">
     work with file-system directories
 
-    <include filename = "../license.xml" />
+    <include filename = "license.xml" />
 
     <constructor>
         Create a new directory item that loads in the full tree of the specified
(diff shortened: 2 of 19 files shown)


sappo:
@paddor that's great stuff. Thanks.

I would encourage you though, to use the zeromq style commits in the further:
```
Problem: <describe the minmal problem you're having>
Solution: <explain how you solved your problem>
```
If you need an example check the git log there are lots of them ;)

paddor:
@sappo You're welcome. :sunglasses: 

Sure, I'll use zeromq style commits from now on.

And thanks for adding me to the teams! Does that mean I don't need to create pull requests anymore? :blush:

sappo:
@paddor you should read our code contract http://rfc.zeromq.org/spec:22

There are two basic rules though, we NEVER commit directly to upstream master. And we NEVER merge our own pull request.

paddor:
@sappo That makes a lot of sense. Thanks again.

sappo:
You are welcome!
Am 20.11.2015 5:25 nachm. schrieb "Patrik Wenger" <notifications@github.com

hintjens:
@paddor the zsock* scripts are kind of confusing now.

I'd like to clean these up. At the least, these things bother me:

* we seem to be generating both XML and C/H for the V3 socket options API.
* we are using 'zsock_option' for the old generated code
* we are using 'zsock_options' for the new generated code (via api/)
* we have code generators in the api directory (they should all be in the same place)

Do you intend to clean this up still or shall I do it?

PR #1174, opened 2015-11-24
problem: polling on multiple thread safe sockets with zpoller is not possible

somdoron:
Solution: make the zpoller work with the new zmq_poller.

@taotetek you can use zpoller with server and client now...


sappo:
I did it. I merged faster than @c-rack :-D

c-rack:
:smile: :+1:

PR #1176, opened 2015-11-24
Problem: zpoller test are failing

somdoron:
Had a bug in the test

### DIFF
--- a/src/zpoller.c
+++ b/src/zpoller.c
@@ -501,7 +501,7 @@ zpoller_test (bool verbose)
     assert (client);
     zsock_t *server = zsock_new (ZMQ_SERVER);
     assert (server);
-    poller = zpoller_new (client, server);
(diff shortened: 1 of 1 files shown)


somdoron:
can be merged now.

sappo:
@somdoron zpoller is failing on my machine with:
```
lt-czmq_selftest: src/zpoller.c:515: zpoller_test: Assertion `which == server' failed.
 * zpoller: make[2]: *** [check-local] Aborted (core dumped)
```
I tried to figure out why and the only thing I could observe was that the zpoller_wait function in line 514 for whatever reason was never entered.

sappo:
Nevermind, forgot to update libzmq to latest version

somdoron:
Ok, anyway I will change the macro of the test to check that the zmq_poller exist (it is now checking if thread safe supported)

PR #1184, opened 2015-11-26
Problem: zsock API has redundant code and other issues

hintjens:
- delete GSL scripts in API, they belong in src
- do not generate api/zsock.xml, this should be made by hand
- do not generate include/zsock_option.h from src/zsock_option.gsl
- api/zsock_options.xml should be called api/zsock_option.xml (no 's')
- generate api/zsock_option.xml from src/zsock_option.gsl

I've probably broken something. However this is the right design for
the files IMO.


paddor:
@hintjens Thank you. Yeah I guess it was needed.

Does that mean you also prefer not to generate api/zsock_option.xml from src/sockopts.xml?

hintjens:
Generating api/zsock_option.xml is a good approach, I think... it means we can wrap the zsock_option API in bindings automatically. I'm not sure how to do the ZeroMQ versioning, or if we need to.

paddor:
Oh sorry, now I see it. You moved it into the src/zsock_option.gsl script. Very nice.

Well, since api/zsock_option.xml is very new (didn't exist before 64957a63271dca435e46578478244b0b5921466e from Nov 19), I guess we can leave loop `.for version where major = "4"` in src/zsock_option.gsl as it is, for now.

hintjens:
Right, though this means that zsock_option.h will no longer support libzmq 3.x and older. Let's leave that as a known issue and fix it later when this is stable.

jemc:
@hintjens If you're thinking of the same thing I'm thinking of (which I'm not sure you are) - my thought earlier was that we can declare all socket options functions regardless of libzmq version (no `#if` macros in the header), but that the functions for which libzmq version was too old would return an error or have no effect (keep the `#if` macros in the source files).

PR #1185, opened 2015-11-26
renamed ZARMOUR_MODE_BASE16_HEX to ZARMOUR_MODE_BASE16

vanatka:
ZARMOUR_MODE_BASE16_HEX because of everywhere was used ZARMOUR_MODE_BASE16


paddor:
That is a generated interface. Fixing it in zarmour.xml...

PR #1192, opened 2015-11-27
Problem: no API XML definition for zconfig

paddor:
Solution: define it

I hope those `CHECK_PRINTF (n)` things are something good. Otherwise, the newly generated header looks functionally identical to me (except for the new function `zconfig_test()`, which does exist in the C file and is mandatory anyway).


paddor:
There's a problem with some of the methods. I declared some of them as fresh, but should only have declared them as mutable...

Comparing with C source and working on a fix...

PR #1198, opened 2015-11-30
Fixes to API files

hintjens:


hintjens:
Oops, it broke something...

PR #1201, opened 2015-12-01
Problem: zsock_option.xml is not installed

sappo:
Solution: use latest zproject to install this file as well


hintjens:
Did you modify src/Makemodule.am by hand, or regenerate it from project.xml? This commit seems incomplete...

sappo:
I regenerated it from project.xml.

hintjens:
I realized after I commented. :)

sappo:
I think I pushed the commit to zproject while you where already on the merge page. In thus it sneaked by you.

PR #1203, opened 2015-12-02
Problem: zconfig_execute() doesn't fail early

paddor:
Even if the handler returns -1, it still recursively processes every
first child of the current zconfig item.

Solution: Fail early as soon as a handler returned -1.

I'm **NOT experienced** in C. Maybe this is completely wrong. But IMHO this is correct (and it makes my CZTop Ruby binding test suite pass).


c-rack:
Looks good to me. Just a small nitpick: even if the comment says otherwise, I would have returned -1 instead of 0, indicating an error.

paddor:
Sure, I'll change it.

Issue #1204, opened 2015-12-02
Problem: in jni binding, there is no task to publish artifacts locally


awynne:
Solution: add the following configuration block to build.gradle (something like this should be done for all projects that have jni bindings, actually)

```
publishing {
    publications {
        jniJar(MavenPublication) {
            artifacts {
                artifact "build/libs/${project.name}-${version}.jar"
            }
        }
    }   
}
```

c-rack:
Thanks, could you send a pull request please?

hintjens:
I'll do this (has to go into zproject).

c-rack:
A warning header in `gradle.build` would help others to find the right direction. I have opened https://github.com/zeromq/zproject/pull/349 to fix this.

c-rack:
@awynne Fixed by https://github.com/zeromq/czmq/commit/89059fb50f5952a7c1da049f840d3c3262a19edb

hintjens:
@awynne I've added that rule and it seems to break things. I've no experience with gradle, to fix this. Could you make it work by hand, and then I'll backport those fixes into zproject. We can adopt this approach for any similar changes (don't be put off by the warning header, just if someone regenerates the project, the changes will be lost).

awynne:
submitted a handwritten fix via a pull request.  @hintjens pls backport as you suggested.

hintjens:
Thanks!

PR #1207, opened 2015-12-02
solution: added ability to publish to local maven repo

awynne:
handwritten fix for #1204 as requested by @hintjens


hintjens:
So this produces the following error, when I run 'gradle build jar':

''''
What went wrong:
A problem occurred evaluating root project 'czmq-jni'.
> Could not find method publishing() for arguments [build_22j1v6ka5m3bsbvjj2ui5s3k78$_run_closure2@3d9a4a3d] on root
''''

Issue #949, opened 2015-03-08
zmsg_load API definition cannot be complete


twhittock:
zmsg_load looks like a 'class member' of zmsg, but can also be a constructor if passed a NULL 1st arg.

This causes issues with definining it using the class API - is the return value 'fresh' or not? It depends on whether the first arg is null (therefore fresh) or a valid zmsg (therefore a reference to the incoming parameter)

It'd maybe make more sense to split it into 2 functions - a new_from_file (or something) and a read to append messages to the existing zmsg. One other thing to note about this method - it silently destroys the input zmsg if there are no frames in the message after loading the file, potentially leaving client code with an invalid reference, if it was passed from above.

jemc:
+1 For having an API where the "freshness" of a return value is statically known.

hintjens:
I'd like to deprecate this method entirely, along with zmsg_save.

twhittock:
OK, should it be deprecated now? What would it be deprecated in favor of?

hintjens:
None of our projects use it. The load/save methods are dubious. The whole
zmsg class is dubious, in fact, and I'd like to start moving CZMQ towards a
single-part model, over time.

jemc:
While we're on the subject, having a way to mark a method as deprecated in the API model might be nice, so that various consumers of the model could respond appropriately, even if its as simple as appending boilerplate deprecation text to the method description.  I may work on this a little later in the week if no one else does.

hintjens:
I'd like to be able to mark every method and global in the API as draft,
stable, or legacy. We don't have raw or retired APIs, I think (raw means no
implementations, retired means no longer implemented).

PR #1220, opened 2015-12-07
Problem: zmsg_load marked as singleton in api model

vyskocilm:
Solution: remove singleton flag and first parameter (implicit for non
singleton methods).

### DIFF
--- a/api/zmsg.xml
+++ b/api/zmsg.xml
@@ -190,11 +190,10 @@
         <return type = "integer" />
     </method>
 
-    <method name = "load" singleton = "1">
+    <method name = "load">
         Load/append an open file into message, create new message if
         null message provided. Returns NULL if the message could not
         be loaded.
-        <argument name = "self" type = "zmsg" />
         <argument name = "file" type = "FILE" />
         <return type = "zmsg" fresh = "1" />
     </method>


jemc:
@vyskocilm - This doesn't seem correct.

The method descriptions says:
> Load/append an open file into message, create new message if null message provided. Returns NULL if the message could not be loaded.

Your change makes it impossible to use the "create new message if null message provided" behavior.

I suggest reverting this.

hintjens:
Do we always assert that self is not null then?

jemc:
@hintjens - Not sure I understand your question.

No, we're not asserting self in `zmsg_load` - it behaves as described and creates a new zmsg in that case - https://github.com/zeromq/czmq/blob/d2b7b08469da4cb11b1afbd824df4cffdd4ee9d9/src/zmsg.c#L616-L617

vyskocilm:
OK, I see - API is quite confusing. Then yes, please revert it ...

hintjens:
So I don't understand why making it a non-singleton stops one passing null
as self...

paddor:
This function does too much IMO. Either way the return value is either fresh or non-fresh, which makes the API wrong in some cases.

jemc:
> So I don't understand why making it a non-singleton stops one passing null as self...

Most object-oriented languages don't allow the `self` argument to be explicitly passed at all.  It's just always the receiver of the method.  So in this case, you'd have to call the equivalent of the pseudocode `NULL.load(file)`, which doesn't work because `NULL` doesn't have a `load` method because it isn't an instance of the `zmsg` class.

So, in almost all object-oriented language, `self`/`this` must always be an instance of the class, and you can't call it any other way.  It may work in the Python bindings, but Python is a bit of an anomaly in this way (letting you pass in an arbitrary `self` value).

A singleton method is not associated with any specific instance, so the "self" argument is an explicit value passed in like any other argument - it is not special like it is for an instance method.

> This function does too much IMO. Either way the return value is either fresh or non-fresh, which makes the API wrong in some cases.

This is an excellent point.

In light of this point, rather than reverting this patch I think we should deprecate calling `zmsg_load` with a `NULL` self (though we probably shouldn't break compatibility by removing it), and create a separate method that *always* creates a new (fresh) `zmsg`.

hintjens:
Yes, I agree. I'd be fine with breaking the API here, and removing the self argument entirely. It is fuzzy and inconsistent with all other methods (self cannot be null, by convention).

hintjens:
In the name of science, I propose we deliberately break the API on this method and see if anyone complains. :-) 

I searched github and found this method used mainly in the titanic pattern (from the Guide).

jemc:
Okay by me.  I'd suggest an API like this:

```
    <constructor name = "load new">
        Load/append an open file into a new message.
        Returns NULL if the message could not be loaded.
        <argument name = "file" type = "FILE" />
    </constructor>

    <method name = "load">
        Load/append an open file into the message.
        Returns NULL if the message could not be loaded.
        <argument name = "file" type = "FILE" />
        <return type = "zmsg" />
    </method>
```

If we agree, I can take care of it in a day or so if no one else beats me to it.

hintjens:
I'd be more brutal: no-one needs the "load into existing message" functionality, it's bogus and breaks the API conventions:

    <constructor name = "load">
        Load/append an open file into a new message.
        Returns NULL if the message could not be loaded.
        <argument name = "file" type = "FILE" />
    </constructor>

twhittock:
Please also see #949

hintjens:
Anyone know of a reason other constructors (e.g. zframe_new_empty) are defined as singletons instead of constructors?

jemc:
@hintjens - No reason I can see.  You're right, they (`zframe`'s `new empty`, `from`, and `recv` methods) should be defined as `constructor`s by convention. A `constructor` implies a `singleton` that returns a `fresh` instance, but we should call them `constructor`s wherever it fits this pattern, just for clarity..

hintjens:
OK, cleaning this up now.

paddor:
I think I had this idea before but then stopped for some reason... I think it was something about one of the bindings... Oh right! I think the python bindings can't handle it. Only one `__init__` possible.

paddor:
I think I even had the python binding fixed (I made only the constructor named `new` become __init__). But there was another thing as well... so I didn't proceed.

hintjens:
@paddor we can handle this by only treating the first constructor as __init__ and the rest as methods... we should not expose this to the API model IMO.

jemc:
Agreed, we should patch the python bindings to be able to deal with more than one constructor.

paddor:
@hintjens True. I hope you'll hit no other problem.

hintjens:
Well, let's force this through. :-)

paddor:
I remember the problem now. How do you do it in Java? A constructor in Java always has the same name as the class, right? Or would you just implement it as a class method? Can't depend on the fact that all constructors have a different signature...

jemc:
@paddor - Yes, we would do it as a singleton method in Java and Python, just as we do in Ruby (where you can only have one constructor called `new`).

However, it makes sense to keep those implementation details out of the API model, as these are all conceptually constructors, and some language bindings will actually support multiple constructors.

hintjens:
I guess for Java I'll only use the 'new' constructor and treat the others as methods.

paddor:
Sure, I love the idea. Just wanted to warn about the problems. It's probably no problem for someone who actually knows Java/Python. I don't. :sweat_smile:

hintjens:
I'll deal with the JNI binding now. Am first fixing all the class files, there's a bunch of these...

twhittock:
I did a lot of these singleton-as-constructor things - I wasn't aware zproject could do more than one constructor. No other reason. Hence the Python bindings' lack of support for the construct, also.

hintjens:
@twhittock makes sense, thanks. As a general rule, when/if we see semantics in the API model that can be improved, we should take the pain now. E.g. methods that pass data back via mutable arguments, or by reference (ghastly IMO).

hintjens:
OK, fixed in zproject_class_api.gsl: https://github.com/zeromq/zproject/pull/368

PR #1221, opened 2015-12-07
Improvements to JNI binding

hintjens:
- return buffers as byte [], not pointers
- add Zstr.str () method

### DIFF
--- a/api/zframe.xml
+++ b/api/zframe.xml
@@ -61,7 +61,7 @@
 
     <method name = "data">
         Return address of frame data
-        <return type = "buffer" mutable = "1" />
+        <return type = "buffer" mutable = "1" size = ".size" />
     </method>
 
     <method name = "dup">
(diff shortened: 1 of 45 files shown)


hintjens:
There is still a warning when building the JNI code and I cannot see what it is:

'''
CZMQ_EXPORT char *
    zstr_str (void *source);

JNIEXPORT jstring JNICALL
Java_org_zeromq_czmq_Zstr__1_1str (JNIEnv *env, jclass c, jlong source)
{
    char *str_ = (char *) zstr_str ((void *) source);
    jstring return_string_ = (*env)->NewStringUTF (env, str_);
    zstr_free (&str_);
    return return_string_;
}

org_zeromq_czmq_Zstr.c: In function ‘Java_org_zeromq_czmq_Zstr__1_1str’:
org_zeromq_czmq_Zstr.c:70:18: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
     char *str_ = (char *) zstr_str ((void *) source);
                  ^
'''

PR #1224, opened 2015-12-08
Problem: doc for zsock_brecv is inaccurate

### LINKED ISSUE #1223: Problem: zsock_brecv documentation unclear
The documentation for zsock_brecv states that none of the returned pointers SHALL be freeed or destroyed. Looking at the code this is not the case for longstr, chunk, msg, frame and uuid.

Is this correct or did I miss something?

hintjens:
Returned objects are owned by caller, who must destroy them.

Solution: fix this.

Fixes #1223

### DIFF
--- a/api/zsock.xml
+++ b/api/zsock.xml
@@ -280,8 +280,11 @@
         reduce memory allocations. The pattern argument is a string that defines
         the type of each argument. See zsock_bsend for the supported argument
         types. All arguments must be pointers; this call sets them to point to
-        values held on a per-socket basis. Do not modify or destroy the returned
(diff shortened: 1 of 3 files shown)


sappo:
That's not true for strings (picture: 's') though, see  https://github.com/zeromq/czmq/blob/master/src/zsock.c#L1388.

They're provided by some socket cache mechanism if I understood it correctly.

hintjens:
Right, strings are so common that allocating/freeing them is costly. So it
stores a single block in the socket object and returns pointers into this
block. This is worth stating in the doc.

PR #1227, opened 2015-12-09
Problem: czmq_selftest block-local variable fails for MSVC

jeffdaily:
A printf statement occurred before the block-local variable. MSVC
failed to compile. Swapping the order fixes the problem.


c-rack:
Please note that the file you have modified is auto-generated using zproject as stated in the file header:
https://github.com/zeromq/zproject

I will fix it there.

PR #1242, opened 2015-12-14
Working on JNI Android builds

hintjens:


jemc:
@hintjens - If you've found the `qt-android` build to be useful outside the context of Qt, then you can feel free to rename it to something more appropriate.  I've always suspected that it would be compatible with other JNI efforts, but never fully confirmed it.

hintjens:
I think we can kill the old android build and rename qt-android...

jemc:
:+1:, since no one seems to be maintaining the other one anyway

hintjens:
Right, will do that then.

Issue #1244, opened 2015-12-14
zcert_save() doesn't fail with empty string as filename


paddor:
During testing of cztop, I realized that zcert_save() doesn't fail when given the empty string as filename. It just creates one file (insted of two): "_secret".

Shouldn't this return an error code?

c-rack:
Good catch. It should either return an error code or `assert(*filename != '\0')` at the beginning:
https://github.com/zeromq/czmq/blob/master/src/zcert.c#L276-L290

c-rack:
Or we just check the return code of the `zcert_save_public` call and return it if not equal to zero:
https://github.com/zeromq/czmq/blob/master/src/zcert.c#L283

I have not tried it, but I'd guess that trying to save a file with empty filename will fail.

paddor:
Okay. I'll try.

paddor:
Yes, zconfig_save() does return -1 if filename is empty.

c-rack:
Closed by #1245

PR #1245, opened 2015-12-14
Problem: zcert_save() ignores result of zcert_save_public()

paddor:
This leads to being able to give the empty string as filename and it'll
create a single file "_secret".

Solution: check zcert_save_public()'s return value

This is for #1244.

### DIFF
--- a/src/zcert.c
+++ b/src/zcert.c
@@ -280,12 +280,13 @@ zcert_save (zcert_t *self, const char *filename)
     assert (filename);
 
     //  Save public certificate using specified filename
-    zcert_save_public (self, filename);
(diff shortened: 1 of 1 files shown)


c-rack:
If you'd have written "This fixes #1244", that issue would have been automatically closed :-)
Good work, anyway!

paddor:
In the pull request or in the commit message? I never know for sure.

c-rack:
Usually, I include it in the commit message, but I think it also works in the pull request message:
https://help.github.com/articles/closing-issues-via-commit-messages/

paddor:
Great! Thanks for the heads up. I'll (try to) remember.

PR #1256, opened 2015-12-16
Fixing VS20xx generation

hintjens:


evoskuil:
The `.vxproj` files are greatly simplified (relative to a typical MSVC project) by factoring out common property sets into a set of common `.props` files. Without these none of the current builds will function. The common properties files should never change, so they either don't need to be generated, or they are simple to generate. I don't have time to go in and add generate for all of these at the moment, just wanted to make you aware of the break.

The VS template that I'm describing is documented here: https://github.com/evoskuil/visual-studio-template

hintjens:
I'd like to generate these props files then (the point is, when we use zproject in a new project, everything should work out of the box).

hintjens:
So we have (a) the properties directory and the files it contains, and (b)
the .props file for the project and for the test? How about the
import.props files, what's their life-cycle?

evoskuil:
* In my reduction the `.vxproj` files are almost entirely boilerplate, apart from the source file listings, so sort of like `Makefile.am`.
* The `.sln` files reference a set of project files.
* The `.filters` files are just editor source file organization (into virtual folders).

(a) The properties directory is reference data, it should never change.
(b) The project-specific `.props` files are where project-internal configuration belongs.
(c) The `.import.props` files are configuration *exports* of dependencies (which are *imports* into this project), so sort of like package config.

The files `libsodium.import.props` and `libzmq.import.props` are copied from those repos. The package config-like registry for these on Windows is **NuGet**, and the imports files are similar to what I place in the NuGet packaging, but these allow one to use the projects directly without a lot of manual config. There is also an import file `czmq.import.props` for the local project, provided for the test project and for depending projects.

The `.xml` files are GUI generators that are pulled in by the corresponding `.props` files. These allow the dev to select configurable options and linkage in the UI. So sort of like `autoconfig`.

So `imports` and related `xml` should change with project configuration options.

You can probably get away without the `resource.h` file, which is source generated in the resource editor. But it's only generated during editing, not by compile. And it's sort of annoying in the editor if it's missing - especially when editing the `.rc` file, which is the source of the binary file info that Windows displays.

The batch files are some scripts that I provide to allow easy execution of all configured builds at the command line. These don't ever need to change.

hintjens:
Thanks for this explanation. I'll aim to backport all this into the
zproject framework. I'll probably have other questions later.

hintjens:
@evoskuil I've back-ported most of this to zproject, and rebuilt the MSVC build tree in CZMQ. There is one point I'm unsure of: how/when to copy the import files from dependent projects. 

My idea is to copy these automagically when running zproject, on the assumption that we check-out projects sanely (as siblings).

evoskuil:
It's a problem that I haven't solved, and your approach seems better than what I've been doing.

I have a set of tools in a private repo that among other things, clone these artifacts into the depending repos. The problem is doing it at a reasonable point so as to keep versions coordinated. I've generally punted that problem, choosing to update these at whatever point I need to update one or more of the builds for whatever reason.

If you can control these updates in zproject by coordinating across repos and versions it should be ideal. For sodium I assume you'd want to peg to particular releases.

evoskuil:
A couple of additional recommendations:

The VS solution, project and filter files are in a compatible format in VS 2910/12/13/15 (not 2008). I use script to generate the set from a master, in each of my repos.

There may be incompatibilities, but my approach uses a fixed subset of features that work in VS 2010 files. The necessary deviations are handled in generation. The schema version must match the target environment (version of VS for .sln files and compiler for .vxproj files). The .props and .xml files don't need this differentiation.

So I recommend you do the same in zproject. I once offered to do this, but never got to it as I haven't learned zproject.

I also recommend you drop the VS 2008 builds as soon as possible, since they are a totally different animal.

hintjens:
So in zproject, the VS 2010-2015 builds are all produced from the same template (used to be 4 templates in zproject, which was not ideal.)

Pegging to sodium releases seems wise. Master seems not always too stable.

PR #1263, opened 2015-12-18
Problem: check-py build fails due to syntax error

c-rack:
See https://travis-ci.org/zeromq/czmq/builds/97634214

Solution: regenerate with latest zproject

### DIFF
--- a/bindings/python/czmq.py
+++ b/bindings/python/czmq.py
@@ -501,7 +501,7 @@ class Zcert(object):
             self.allow_destruct = args[1] # This is a 'fresh' value, owned by us
         else:
             assert(len(args) == 2)
-            self._as_parameter_ = lib.zcert_new from(args[0], args[1]) # Creation of new raw type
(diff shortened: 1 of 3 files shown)


hintjens:
There still seems to be an error in the Python build:

```
/bin/bash ./libtool --mode=execute -dlopen src/libczmq.la python bindings/python/test.py
Traceback (most recent call last):
  File "_ctypes/callbacks.c", line 314, in 'calling callback function'
  File "bindings/python/test.py", line 1052, in _echo_actor
    msg = Zmsg.recv(pipe)
AttributeError: type object 'Zmsg' has no attribute 'recv'
make: *** [check-py] Segmentation fault (core dumped)
```

c-rack:
Yes, saw it, and it seems that your assumption in 130ef018170d5e16bb4f38673a580ad2a2c20026 was correct :-)
Hope to find time to investigate further over the weekend...

hintjens:
Ah, makes sense.

PR #1266, opened 2015-12-21
android jni build.sh: ensure "gradle build jar" is run

zoobab:
android jni build.sh: ensure "gradle build jar" is run before, otherwise compilation will fail with missing headers.


hintjens:
@zoobab this seems to be just a whitespace fix? The old and new code are the same.

c-rack:
No, he commented-in two lines by removing the `#`.
The contrast of gray on light-red is not very readable.

hintjens:
@zoobab, my bad. I'd commented these two lines out of the generated code to
make things run faster, then committed it by mistake. Thanks for fixing it!

PR #1277, opened 2015-12-26
Problem: repetitive need to delete non-tracked files after generating

paddor:
Or, in case of the .vimsession file, I just don't want git to consider
it dirt.

Solution: gitignore them

Since those other files in the build directory aren't tracked, I assume it's okay to gitignore them.


hintjens:
You can add any generally useless files to zproject/zproject_git.gsl, if you like.

paddor:
@hintjens Thanks for pointing that out.

PR #1301, opened 2016-01-13
Problem: zclock does not have api model

vyskocilm:
Solution: make one using mkapi.py


vyskocilm:
The second time and I forgot to remove the comment - this is anoying despite I wrote the code :-|
Thanks for spotting it!

https://github.com/zeromq/czmq/pull/1302

PR #1309, opened 2016-01-18
Revert "I don't think class status belongs in project.xml"

hintjens:
Reverts zeromq/czmq#1308

Oops.

I just realized we don't have xml specs for the older classes.

### DIFF
--- a/project.xml
+++ b/project.xml
@@ -9,44 +9,44 @@
     <use project = "libzmq" />
     <use project = "uuid" optional = "1" implied = "1" />
 
-    <class name = "zactor" />
-    <class name = "zauth" />
-    <class name = "zarmour" />
-    <class name = "zbeacon" />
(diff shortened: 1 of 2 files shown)


bluca:
I should have clarified in the commit message why I put the in the project.xml, sorry about that!

Note that, although more confusing IMHO, having the state defined in project.xml for some classes and in api/CLASS.xml for others will work just fine. Also note if defined in both with different values, the value in project.xml wins (by accident rather than by design - haven't looked into how to change it).

hintjens:
It's all good, I'm almost done. We'll have the state at various levels. In
some cases I want to specify it at the project (top item) level.

I've done a bunch of changes. One thing I'm not sure how we do is disable
drafts in stable releases automatically. Perhaps we can patch platform.h
during "make dist"?

PR #1315, opened 2016-01-21
draft related improvements

paddor:
* socket options can be draft now (like heartbeat options)
* mark a few unreleased, recent API functions as draft
* regenerate Ruby binding

This allows CZTop to be used with the current stable release of CZMQ.


hintjens:
@paddor fantastic!

paddor:
Thank you. :)

PR #1316, opened 2016-01-22
Problem: czmq_selftest failing on zsys

hintjens:
We used to run zsys_test first, so it could test set_io_threads ()
which has to be the first thing a process does before creating any
sockets.

With zproject's test framework, we don't control the order of test
cases (and this is good, they should not be dependent). Now zsys is
called late, and so the call to set_io_threads () asserts.

Solution: remove this specific test.


hintjens:
<blinks> and it's merged! Thanks :)

sappo:
Merging in less than 30s is fun :)

PR #1321, opened 2016-01-24
Problem: selftest build must be optional

### LINKED ISSUE #1320: Question: building without self_test.
Is it no longer possible to build without `czmq_selftest`?

I see `--without-zmakecert` has been changed to `--disable-zmakecert` but `--without-czmq_selftest` has been removed.

hintjens:
Solution: fixed in zproject, regenerated here.

Fixes #1320


evoskuil:
This restores the configure option but the build is not conditional: https://github.com/zeromq/czmq/blob/master/src/Makemodule.am#L151-L155

hintjens:
OK, let me check that.

PR #1322, opened 2016-01-24
Problem: czmq_selftest build still not conditional

hintjens:
Solution: fixed in zproject and rebuilt here.


evoskuil:
Thanks, this worked around my linking problem. I'll create a new issue for that.

PR #1325, opened 2016-01-26
Problem: CZMQ doesn't build with MSVC without DRAFT_API definition

keent:
Solution: Add DRAFT_API defs


sappo:
@keent is msvc capable of setting a define using some kind of switch. Because your solution is only a workaround to must be manually applied to all zproject based projects. It would be nice to solve this issue in zproject itself.

keent:
@sappo I will try to learn zproject. I talk to Pieter about this and he said - for now to apply the fix by hand and he'll backport it.

PR #1326, opened 2016-01-27
Problem: auto-generated python cffi-bindings do not exist

asokoloski:
pyczmq is maintained by hand but apparently this is painful and
therefore sporadic.  This binding doesn't completely replace it but
it's a starting point.  Also it was easy to do since I wanted to
generate cffi bindings for malamute anyway.


asokoloski:
thanks :)

PR #1339, opened 2016-02-02
Problem: czmq is out-of-sync with latest zproject

c-rack:
Solution: re-generate with latest zproject

Fixes #1338

PR #1340, opened 2016-02-02
Problem: JNI bindings were out of date

hintjens:
Solution: regenerate via zproject


c-rack:
Maybe same as #1339

hintjens:
Merged it myself just out of curiosity...

PR #1342, opened 2016-02-03
Problem: can't configure Windows builds

hintjens:
There are a bunch of problems emerging:

- can't enable/disable drafts explicitly
- can't make draft decision automatically
- can't detect presence/absence of packages

Solution: add 'configure' step for MSVC

This is builds/msvc/configure.bat. It looks for dependencies in a
sibling directory. It also checks if we're building from git master
or from a source package.

Then it generates platform.h accordingly.

Relatedly, the build.bat script passes parameters to msbuild so that
we skip any optional packages that aren't present.

On MSVC, "package is present" simply means the source tree exists
where we expect it, and hopefully has been built.

I've also made the same changes in libzmq, so one can configure
it with / without libsodium.

Still to do:

- enable/disable building mains        --enable-xyz
- without packages, even if present    --without-xyz


hintjens:
Note the real changes are in zproject.

PR #1350, opened 2016-02-08
Support of pre-allocated FDs

bluca:
Add pre_allocated_fd sockopt to sockopts.xml and regenerate.

### DIFF
--- a/api/zsock_option.xml
+++ b/api/zsock_option.xml
@@ -38,6 +38,16 @@
     <argument name = "heartbeat timeout" type = "integer" />
 </method>
 
+<method name = "pre allocated fd" polymorphic = "1" state = "draft">
+    Get socket option `pre_allocated_fd`.
+    <return type = "integer" fresh = "1" />
+</method>
+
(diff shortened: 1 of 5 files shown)


hintjens:
I know this is a little late, yet the name 'pre_allocated_fd" seems clumsy and could be simpler.  I should make this comment on libzmq since it affects that too. I know a lot of the recent ZMQ options have long names. It's not a trend we should be encouraging IMO. I'll be patching some of the worst.

So how about ZMQ_PREFD? or ZMQ_USEFD?

(Put it like this: the shorter and sweeter the name, the more chance people will use the function and the less chance some slash-and-burn artist like me will rip it out in a future version.)

bluca:
Sure, I don't have strong opinions on the name.

I picked that one because there is already a ZMQ_FD for a completely different purpose, so I wanted to avoid confusion and make it clear, but we have documentation for that.

Unless someone has a better name, tomorrow I'll push PRs to change to ZMQ_USEFD to both repositories.

PR #1360, opened 2016-02-18
Fixes for Windows

hintjens:


include/czmq_prelude.h
```diff
@@ -460,11 +460,10 @@ typedef struct {
     typedef unsigned int  uint;
 #   if (!defined (__MINGW32__))
     typedef int mode_t;
-#       if defined (__IS_64BIT__)
-    typedef long long ssize_t;
```
bluca: Are these intended to be removed?

PR #1368, opened 2016-03-04
Problem: zauth_v2 sometimes hangs in self test

lerwys:
This is a backport to the zauth hanging problem
described in commit zeromq/czmq@b6805df78 by hintjens.


hintjens:
Nice...

PR #1370, opened 2016-03-10
Problem: Build instructions are out of date in README.md

opedroso:
Solution: Update build instructions for both Linux and Windows in README.md

To the person doing the merge: if it does not make sense to modify .gitattributes file because it will affect all users, please skip it and only apply the README.md changes. Thanks!


sappo:
Closing this as the changes have been accepted through other PRs

PR #1385, opened 2016-04-08
Update README.md, missing code block start

saulthu:


c-rack:
Please note that README.md is auto-generated using README.txt using `gitdown`, so next time your change might be overwritten.

PR #1389, opened 2016-04-12
ZFrame metadata getter

petehayes102:
This is a tiny PR, but I can't see another way to get access to frame metadata.

### DIFF
--- a/src/zframe.c
+++ b/src/zframe.c
@@ -223,6 +223,19 @@ zframe_data (zframe_t *self)
     return (byte *) zmq_msg_data (&self->zmsg);
 }
 
+//  --------------------------------------------------------------------------
+//  Return meta data property for frame.
+//  Caller must free string when finished with it.
+
+const char *
+zframe_meta (zframe_t *self, const char *property)
+{
+    assert (self);
+    assert (zframe_is (self));
+
+    return zmq_msg_gets (&self->zmsg, property);
+}
+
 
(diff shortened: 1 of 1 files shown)


sappo:
@petehayes102  could you please in short state what's your problem and how this PR solves it. Thanks!

petehayes102:
Sure. I have an app that uses ZAuth to authenticate clients. I want to be able to access message meta data without having to directly use a ZMQ message (i.e. use CZMQ structs instead). ZFrame uses zmq_msg_t under the hood, so this is a good place to retrieve metadata, e.g. user ID, via zmq_msg_gets().

My PR provides a CZMQ wrapper to get message meta data from a ZFrame, which currently doesn't seem possible in CZMQ.

sappo:
Okay, thanks!

For further contributions it would be great if you describe the Problem you're having and the Solution in your commit message really short. Have a look at the commit log for some examples if you need any. This usually speeds up merging of a PR.

petehayes102:
Sure. Sorry about that!

bluca:
Is this intended to be a public API?

If so, it should be added to the headers via api/zframe.api and regeneration through zproject. And also it should be unit-tested.

petehayes102:
Ok, I didn't know about the api thing. I'll raise another PR with the missing bits.

bluca:
Thanks!

PR #1390, opened 2016-04-12
Added docs and unit test for zframe_meta()

petehayes102:
Added missing unit test and docs from #1389.

### DIFF
--- a/src/zframe.c
+++ b/src/zframe.c
@@ -571,9 +571,9 @@ zframe_test (bool verbose)
 
     //  @selftest
     //  Create two PAIR sockets and connect over inproc
-    zsock_t *output = zsock_new_pair ("@inproc://zframe.test");
+    zsock_t *output = zsock_new_pair ("@tcp://127.0.0.1:9001");
     assert (output);
-    zsock_t *input = zsock_new_pair (">inproc://zframe.test");
+    zsock_t *input = zsock_new_pair (">tcp://127.0.0.1:9001");
     assert (input);
 
     //  Send five different frames, test ZFRAME_MORE
@@ -633,6 +633,16 @@ zframe_test (bool verbose)
         zframe_destroy (&frame);
     }
     assert (frame_nbr == 10);
+
+    // Test zframe_meta
+    frame = zframe_new ("Hello", 5);
+    assert (frame);
+    rc = zframe_send (&frame, output, 0);
+    assert (rc == 0);
+    frame = zframe_recv (input);
+    assert (streq (zframe_meta (frame, "Socket-Type"), "PAIR"));
+    assert (zframe_meta (frame, "nonexistent") == NULL);
+    zframe_destroy (&frame);
 
     zsock_destroy (&input);
     zsock_destroy (&output);
--- a/api/zframe.api
+++ b/api/zframe.api
@@ -61,6 +61,13 @@
         <return type = "buffer" mutable = "1" size = ".size" />
     </method>
 
+    <method name = "meta">
+        Return meta data property for frame
+        Caller must free string when finished with it.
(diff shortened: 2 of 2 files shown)


src/zframe.c
```diff
@@ -571,9 +571,9 @@ zframe_test (bool verbose)
 
     //  @selftest
     //  Create two PAIR sockets and connect over inproc
-    zsock_t *output = zsock_new_pair ("@inproc://zframe.test");
+    zsock_t *output = zsock_new_pair ("@tcp://127.0.0.1:9001");
```
bluca: Why this change?

bluca:
```
src/zframe.c: In function ‘zframe_meta’:  CC     src/src_libczmq_la-zhashx.lo
src/zframe.c:236:5: error: implicit declaration of function ‘zmq_msg_gets’ [-Werror=implicit-function-declaration]
src/zframe.c:236:5: error: return makes pointer from integer without a cast [-Werror]
```

src/zframe.c
petehayes102: Argh...I stuffed that up, sorry. I was going to use zauth to return some meta data I could test, which doesn't work on inproc as far as I know. I didn't end up doing that, but now that I've reverted it to inproc, I'm getting a segfault... I'll need a few minutes to figure it out. I'm not very experienced with C.

petehayes102:
I compiled this on OS X and didn't see those errors. I'll take a look.

src/zframe.c
petehayes102: Ok, I got a segfault because I was running streq() on NULL. Duh!

Metadata doesn't get passed over inproc. Not sure why.

bluca: Ah ok, makes sense, thanks.

petehayes102:
@bluca I just did a build on a fresh CentOS machine and wasn't able to reproduce the errors you got. Could you give me some more info about your platform?

I am compiling against ZMQ 4.1.4. What ZMQ version are you using?

I'm not sure why the compiler is complaining about an "implicit declaration" either. It seems all the right headers are there: zframe.c => czmq_classes.h => czmq.h => czmq_library.h => zmq.h => ZMQ_EXPORT const char *zmq_msg_gets

Any ideas?

bluca:
So if you check on the CI build: https://travis-ci.org/zeromq/czmq/builds/122511912
The builds with older versions of ZMQ are failing. That's because the API you are using was introduced in 4.1. So, your new API should check the version like others are doing before using it, with \#ifdefs and so on. You can find a lot of examples in the code. Would you be able to send another PR with a fix?

I'll merge this in the meanwhile given it's not related to this changes.

petehayes102:
Thanks. Will do.

hintjens:
Yes, metadata isn't passed over inproc because it's generated by
authentication, and inproc connections don't do authentication.

bluca:
2 more comments:

The api file is not only a documentation, is also the definition use by zproject, the code generator, to generate the API headers and stubs. After adding a new API there, the generation should be re-run and the result checked in. You can find more information here: https://github.com/zeromq/zproject

TL;DR: compile and install gsl, install zproject, run gsl project.xml in the czmq tree.

Then, for future contributions, please use the following format for the git commit:

```
Problem: brief one-line description of the problem you are solving

Solution: details of your solution
```

petehayes102:
RE: Commit messages - I'm an idiot. I meant to do it for this PR and then went on autopilot. Sorry about that.

RE: API - Did you want me to PR the new zproject output along with the version gate stuff?

bluca:
No worries!

If you could, that would be great. Feel free to do it separately, and ask for help for zproject if you need (learning curve can be a bit steep).

PR #1391, opened 2016-04-12
Fix issues with previous zframe_meta PRs

petehayes102:
I've added a version gate around zframe_meta and slightly modified the unit test to ensure that it won't segfault.

@bluca I haven't committed the GSL update yet. When I ran `gsl project.xml` it seemed to make a lot of trivial changes to a lot of files (34 in total). I assume that this isn't expected?

### DIFF
--- a/src/zframe.c
+++ b/src/zframe.c
@@ -640,7 +640,9 @@ zframe_test (bool verbose)
     rc = zframe_send (&frame, output, 0);
     assert (rc == 0);
     frame = zframe_recv (input);
-    assert (streq (zframe_meta (frame, "Socket-Type"), "PAIR"));
+    const char *meta = zframe_meta (frame, "Socket-Type");
+    assert (meta != NULL);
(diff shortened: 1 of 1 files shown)


bluca:
Thanks!

Yes, the changes are expected, as zproject will have moved on since the last time it was run. If you want to be more precise, you could just add and commit the changes caused by your new API separately, but it's up to you.

Also, regarding the \#ifdef, the way we handle it usually with an API that needs a certain feature/version from ZMQ, is that the API will always exist, otherwise the build will break depending on the version of libzmq. So the \#ifdef should be inside the function, and the \#else would just be a simple `return NULL;`. This way you can build with any version of the library. Could you please fix this?

petehayes102:
Ok, that makes sense. I'll create another PR. I might set some kind of record?

I'll also commit all of the changes made by GSL.

bluca:
Perfect, thanks!

PR #1395, opened 2016-04-13
Problem: index.js should export constants + native

rgbkrk:
Solution: Export the constants with the native bindings

Running into issues using the bindings as is, though this is a start for how the index can look. We need to figure out what our exported interface on the top-level is going to be here as well.


hintjens:
Nice :)

PR #1397, opened 2016-04-13
Problem: userland needs bindings install

rgbkrk:
Solution: run `npm install --save bindings`

Solely for the node bindings

### DIFF
--- a/bindings/nodejs/package.json
+++ b/bindings/nodejs/package.json
@@ -17,8 +17,9 @@
     "url": ""
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


hintjens:
Nice. I've backported this into zproject/zproject_nodejs.gsl so that it'll cover all projects. This file is generated, though I can't put a warning header in it since there's no space for comments afaic.

rgbkrk:
Ah, right. Thanks for doing that.

Might be able to add a dummy field to the package.json as well.

PR #1398, opened 2016-04-13
Add certificate metadata to ZAP handler reply

petehayes102:
This addresses an existing TODO comment in zauth.


petehayes102:
Looks like the Ubuntu builds (i.e. Travis) are more picky than OS X. I'll create a new PR to address the error. Sorry about that.

PR #1400, opened 2016-04-13
Problem: Testing zauth certificate metadata breaks on ZMQ < 4.1.

petehayes102:
Solution: Add a version gate to prevent running the test on earlier ZMQ versions.

### DIFF
--- a/src/zauth.c
+++ b/src/zauth.c
@@ -672,6 +672,7 @@ zauth_test (bool verbose)
         success = s_can_connect (&server, &client, false);
         assert (success);
 
+#if (ZMQ_VERSION >= ZMQ_MAKE_VERSION (4, 1, 0))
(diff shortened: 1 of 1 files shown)


hintjens:
Thanks for the neat PRs, @petehayes102 :)

petehayes102:
No worries mate. Sorry that it took a few of them to get it right!

bluca:
In case you are not familiar with Travis and you would like to check before opening a PR, you can go on travis-ci.org, login with your Github account and then enable your forks. Then when you push a branch it will get built.
Note that we don't require it, it's perfectly OK if it takes a couple PRs to get it right - it's up to you.

hintjens:
Yes, no worries with iterative changes that take time to smooth out.

petehayes102:
Thanks guys :)

PR #1404, opened 2016-04-18
Problem: UUID not in uppercase

hitstergtd:
Solution: Change 'uuids' to 'UUIDs'


bluca:
Note that README.md is autogenerated, so you'll want to update README.txt as well otherwise the next time zproject is ran it will be overwritten

hitstergtd:
Thanks and understood. I have submitted another pull, #1405, which fixes all such instances.

PR #1409, opened 2016-04-21
problem: czmq doesn't support new scatter/gather socket types

somdoron:


sappo:
interesting, is there an documentation regarding that sockets?

somdoron:
not yet, just created it today on libzmq. It works the same as push/pull but thread safe and single frame.

api/zsock.api
```diff
@@ -126,6 +126,16 @@
         <argument name = "endpoint" type = "string" />
     </constructor>
 
+    <constructor name = "new gather" state = "draft">
+        Create a GATHER socket. Default action is bind.
+        <argument name = "endpoint" type = "string" />
+    </constructor>
+
+    <constructor name = "new scatter" state = "draft">
+        Create a GATHER socket. Default action is connect.
```
sappo: This should be SCATTER instead of GATHER, right?

taotetek:
@somdoron This is wonderful. I'm going to add support for these to my rsyslog plugins when building against a czmq version that has them.

sappo:
That's neat :+1:

api/zsock.api
somdoron: yes, fixing that.

hitstergtd: @sappo Oops I just saw this! I went and pulled @somdoron's latest, should have let you do it. :(

sappo: @hitstergtd not at all, the faster we merge correct PRs the better. Even though I love pushing that green button :P

hitstergtd: @sappo - hah, thanks! I don't get to push that green button often enough, so have itchy trigger finger when it comes to Pulls. :)

bluca:
This is really awesome stuff!

api/zsock.api
sappo: @hitstergtd I can relate ;)

PR #1411, opened 2016-04-21
Allow loading certs from an arbitrary location in zcertstore

petehayes102:
This is a bit of a speculative PR in that I'm not sure whether you'll agree on my implementation. I'm very open to suggestions/discussion.

### DIFF
--- a/api/zcertstore.api
+++ b/api/zcertstore.api
@@ -25,6 +25,16 @@
         stored on disk.
     </destructor>
 
+    <callback_type name = "loader">
+        Loaders retrieve certificates from an arbitrary source.
+        <argument name = "self" type = "zcertstore" />
+    </callback_type>
+
+    <method name = "set_loader">
+      Override the default disk loader with a custom loader fn.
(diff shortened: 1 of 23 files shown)


sappo:
Your implementation seems fine to me. But I couldn't figure out `zcertstore_empty`'s purpose is exactly Maybe you can give me a more concrete example.

bluca:
Nice stuff! Is the old behaviour unchanged? IE, would an application written against the previous version still work? If not, we'll need to add it to the notices at release time.

petehayes102:
@sappo `zcertstore_empty` is really useful for writing bindings without having to create a native struct. For example, in Rust I can treat zcertstore as a raw pointer and don't care about what it represents. Thus I can call `zcertstore_empty` to flush the hash table without having to get the hash table pointer from the zcertstore pointer. Without that method, I'd have to create a struct in Rust to represent zcertstore in order to access its internals.

@bluca Yep I made sure that everything functions normally and doesn't change old behaviour. If you look at my tests, they are at the bottom. All of the original tests are unmodified.

PR #1413, opened 2016-04-24
Give zcertstore loaders ability to store state

petehayes102:
This builds on my previous PR #1411 to allow any zcertstore loader to store state, rather than just providing a callback. This gives us two nice things:

1. It makes the interface consistent between the default disk loader and user-defined loaders
2. It allows user-defined loaders to store, for example, a connector to their certificate store, transaction IDs etc.

### DIFF
--- a/api/zcertstore.api
+++ b/api/zcertstore.api
@@ -30,9 +30,16 @@
         <argument name = "self" type = "zcertstore" />
     </callback_type>
 
+    <callback_type name = "destructor">
+        Destructor for loader state.
+        <argument name = "self_p" type = "buffer" c_type = "void **" />
+    </callback_type>
+
     <method name = "set_loader">
       Override the default disk loader with a custom loader fn.
       <argument name = "loader" type = "zcertstore_loader" callback = "1" />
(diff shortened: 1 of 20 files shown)


bluca:
@petehayes102 - in line with our development model, new APIs should be marked as state = "draft" when first introduced. Could you please send a PR to change loader, set_loader, empty and destructor and regenerate from zproject?

This is useful since, as it happens in this PR, the API is subject to change until declared "stable" and it helps reduce problems for users of the library, who expects "stable" API not to change.

petehayes102:
Oh right. I'll do that now.

bluca:
Thanks!

petehayes102:
@bluca Given that these are now draft APIs, when/how will a decision be made as to their stability? Is that part of the lead up to release?

bluca:
I don't think it has been formalized as a process, but I think it's part that, and also in concert with the developer of the API. When you are happy with the state of it, are using it and finished all the changes you want to do to it can be discussed to move it to the stable state.
This allows to commit and share the code early on in the development process.

PR #1414, opened 2016-04-24
Problem: New zcertstore APIs were not marked as "draft".

petehayes102:
Solution: Mark as "draft".

As discussed in #1413.

### DIFF
--- a/api/zcertstore.api
+++ b/api/zcertstore.api
@@ -25,17 +25,17 @@
         stored on disk.
     </destructor>
 
-    <callback_type name = "loader">
+    <callback_type name = "loader" state = "draft">
         Loaders retrieve certificates from an arbitrary source.
         <argument name = "self" type = "zcertstore" />
     </callback_type>
 
-    <callback_type name = "destructor">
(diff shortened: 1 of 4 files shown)


bluca:
Thanks!

Unfortunately it looks like the draft system causes problems with typedefs. It's not an issue with your changes, but in zproject I think. I'll have a look.

petehayes102:
Damn... Sorry for the bother. Let me know if I can help.

bluca:
No problem, not your fault! This actually helped uncover an issue in zproject :-)
I've got a solution, I'll send a PR shortly.

bluca:
Opened https://github.com/zeromq/zproject/pull/600 to fix in zproject, will regenerate here once merged.

Ah also, for future contributions, could you please have a separate commit with the result of the zproject regen? This allows to cleanly separate in the git history the api/* changes from the zproject regeneration.

petehayes102:
@bluca Yep, will do. One of these days I'll get a handle on the conventions! :)

PR #1415, opened 2016-04-24
problem: czmq doesn't provide timers when using zpoller

somdoron:
we can use zloop but zloop doesn't support new socket type (and it is a pretty big class).

solution: ztimerset, class that can used together with zpoller.

Example:
```C

// Adding new timer, parameters: Interval, handler function, argument
ztimerset_add (set, 10, handler, arg); 

// Get the timeout from the timereset
int timeout = ztimerset_timeout (set);

// Call zpoller_wait with the timeout
zpoller_wait (poller, timeout);
 
// Immediately after wait is called calling execute to invoke all timers
ztimerset_execute (set); 
```

We might want in the future to make a method that do the 3 methods calls in one go. might be ztimerset_wait that accept zpoller or zpoller was provided in the constructor.


bluca:
Nice!

bluca:
@somdoron - it looks like the test fails on stable releases of libzmq:

```
lt-czmq_selftest: src/ztimerset.c:168: ztimerset_test: Assertion `self' failed.
 * ztimerset: make[2]: *** [check-local] Aborted (core dumped)
```

https://travis-ci.org/zeromq/czmq/builds/125367812

Some issue with the ifdef?

somdoron:
Yes, I forgot to avoid to the self test if zmq_timers doesn't exist, fixing this now

somdoron:
#1417 should fix this

PR #1417, opened 2016-04-24
problem: ztimerset failing on stable libzmq release

somdoron:
only run test when building with libzmq from version 4.2

### DIFF
--- a/src/ztimerset.c
+++ b/src/ztimerset.c
@@ -160,6 +160,7 @@ void handler(int timer_id, void* arg)
 void
 ztimerset_test (bool verbose)
 {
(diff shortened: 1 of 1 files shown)


bluca:
:+1:

somdoron:
@bluca  anyway to avoid compile this class completly if not using libzmq > 4.2?

bluca:
Mmh I don't think we have that facility in the build system at the moment. Would be useful indeed. But I guess the problem would be that the public header (and thus API+ABI) would change in a stable build depending on the underlying version of libzmq.

somdoron:
ok thanks

PR #1424, opened 2016-04-26
Problem: zframe_set_group and zframe_group not exported

taotetek:
This pr adds the zframe_group and zframe_set_group calls to the api model for zframe, and updates the zframe header to export these functions.

Please note I did not regenerate all of the bindings via zproject - if that's desired I can do so and send a followup PR (or someone else is welcome to).

### DIFF
--- a/api/zframe.api
+++ b/api/zframe.api
@@ -104,6 +104,17 @@
         <argument name = "more" type = "integer" />
     </method>
 
+    <method name = "group">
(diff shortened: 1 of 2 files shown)


somdoron:
I think they should be marked as draft

somdoron:
Actually the methods already exposed in the api but as draft. We may need to revert?

PR #1425, opened 2016-04-26
Problem: missed argument on zframe_set_group model

taotetek:
I missed the argument in the model for zframe_set_group - fixed!


somdoron:
@taotetek, the method already exposed in the api, new they defined twice...

somdoron:
As draft

taotetek:
@somdoron aha! I haven't done a PR since the new draft stuff was added. Doh!

taotetek:
@somdoron I'll close this one and fix, thanks for the info! I ended up thinking it wasn't exported when trying to link rsyslog against it - I must not be passing the right flags to build w/ draft.

PR #1426, opened 2016-04-26
Problem: zframe group api already included as draft

taotetek:
Solution: remove the added additional definition that is not needed.

### DIFF
--- a/api/zframe.api
+++ b/api/zframe.api
@@ -104,17 +104,6 @@
         <argument name = "more" type = "integer" />
     </method>
 
-    <method name = "group">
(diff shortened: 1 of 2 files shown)


sappo:
@taotetek this PR somehow makes the MacOS build fail. I couldn't make any sense of the failure maybe you can have a look at it when you got the time. 

See: https://travis-ci.org/zeromq/czmq/jobs/125907909#L430

sappo:
@taotetek looking closer at your PRs there must be something off with travis.

sappo:
@taotetek just re-ran the tests now everything is fine *weird*

PR #1431, opened 2016-05-02
Problem: There is no README documentation for ztimerset

sappo:
Solution: Generate and add it to README!


somdoron:
@sappo how do you regenerate everything?

Issue #1434, opened 2016-05-03
PROBLEM: zfile_eof () does not work


karolhrdina:
Minimal code to reproduce here:
https://gist.github.com/karolhrdina/13a23c33c5168e76b061927c0ac7ebea

bluca:
Closing, fixed by https://github.com/zeromq/czmq/pull/1435

PR #1435, opened 2016-05-03
Problem: issue #1434 zfile_eof () doesn't work

malanka:
Solution: check the "eof" condition at the right place


sappo:
@malanka Nice, would you mind adding a test for this method as well, thanks!

malanka:
Sure :)

PR #1437, opened 2016-05-05
Problem: zsys_set_logstream(NULL) not functional.

evoskuil:
This is documented as a means to disable log output to the console, yet it is reset on every call to `s_log()`. This makes the code that follows immediately after the deleted code half dead:

```c
    if (!s_logstream)
        s_logstream = stdout;

    if (s_logstream || s_logsender) {
```

as `s_logstream` can never be `NULL`. This restores rational/documented behavior.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+CZMQ version 3.0.2 stable, released on 2015/06/08
+=================================================
+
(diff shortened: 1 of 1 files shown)


evoskuil:
Oops, inadvertently closed.

PR #1442, opened 2016-05-11
Problem: uuid library not used

bkmit:
Solution: list uuid in project.xml as optional default on if found and regenerate project


bluca:
Hi, I think your local version of zproject is out of date, could you please make sure it has the latest changes? Otherwise this is breaking a few things that were fixed recently. THanks!

bkmit:
You are right @bluca ! Good catch, thank you.

bluca:
Thanks!

For the next contributions, please add the regeneration from zproject in a separate git commit. This helps keep configuration changes more visible, since the regeneration is very noisy usually.

bkmit:
Got it. Will do commit regeneration alone.

PR #1444, opened 2016-05-13
Problem: wrong License and URL Fields in czmq.spec

cyrilleverrier:
Fix License and URL fields


bluca:
Hi,

Note that the spec is autogenerated from zproject, so next time it will be overwritten.

I fixed the autogenerated Debian package stuff a while ago in zproject, so you might want to do it there too to avoid the fixes getting lost.

Thanks!

PR #1446, opened 2016-05-18
Fix/zproxy assertion fail

piskorzj:
This pull request fixes concurrency (or hazard) issues I've observed when working with XPUB/XSUB proxy on distributed system.

Related issue #1445.

Issue #1445, opened 2016-05-17
Undocumented zproxy behavior is causing assertion


piskorzj:
Configuring zproxy actor is not an atomic operation, and received message between setting **FRONTEND** and **BACKEND** is causing assertion in `s_self_switch`.

Here is simple program to reproduce this issue:

```
#include <czmq.h>

int main(int argc, char **argv) {
  zactor_t *proxy = zactor_new (zproxy, NULL);
  assert (proxy);
  
  //  Connect application sockets to proxy
  zsock_t *faucet = zsock_new_push (">inproc://frontend");
  assert (faucet);
  zsock_t *sink = zsock_new_pull (">inproc://backend");
  assert (sink);
  
  zstr_sendx (proxy, "FRONTEND", "PULL", "inproc://frontend", NULL);
  zsock_wait (proxy);
  
  zstr_sendx (faucet, "Hello", "World", NULL);

sleep(1);

  zstr_sendx (proxy, "BACKEND", "PUSH", "inproc://backend", NULL);
  zsock_wait (proxy);

  return 0;
}
```

The easiest solution for this problem is to first configure *output* socket (like PUB, PUSH...) and then *input* ones. However, this is not documented anywhere.

sappo:
Would you like to update the documentation and send a PR?

piskorzj:
Sure, editing zproxy.c is enough?

sappo:
Yes, thanks!

piskorzj:
Problem is not that simple as I thought. Presented *solution* works for unidirectional socket (PUSH/PULL) quite well, however in my use case I need XPUB/XSUB. Now changing order of sockets gives nothing, because SUBSCRIBE also causes assertion.

Here is code snippet to reproduce:
```
#include <czmq.h>

int main(int argc, char **argv) {
  zactor_t *proxy = zactor_new (zproxy, NULL);
  assert (proxy);

  zsock_t *faucet = zsock_new_pub (">inproc://frontend");
  assert (faucet);

  zsock_t *sink = zsock_new_sub (">inproc://backend", "whatever");
  assert (sink);
 
  zstr_sendx (proxy, "BACKEND", "XPUB", "inproc://backend", NULL);
  zsock_wait (proxy);
 
  zstr_sendx (proxy, "FRONTEND", "XSUB", "inproc://frontend", NULL);
  zsock_wait (proxy);
  
  return 0;
}
```

[...]

PR #1447, opened 2016-05-19
Test fail on older platform after fixing zproxy issue

piskorzj:
After #1446 Tests were failing on older zeromq releases because on them connecting before bind in inproc transport is forbidden.

Original issue #1445


bluca:
@piskorzj - Thanks! For your next contributions, could you please follow the guideline for the commit message:

```
Problem: X and Y are broken

Solution: fix X with such and such
fix Y with such and such
```

piskorzj:
Sure, I will!

PR #1449, opened 2016-05-24
Problem: autotools out of sync (zproject has evolved!)

sappo:
Solution: re-generate all!


bluca:
Thanks for fixing this, it fell off my radar :-)

sappo:
You're welcome :)

PR #1454, opened 2016-06-01
Fix compilation with mingw64 using autotools

diorcety:
HI,
Here the fixes for correctly compile czmq using msys2 + mingw64 gcc.
DLL_EXPORT is automatically generated by libtool. I'm using this symbol + the indication of internal compilation put in the configure.ac in order to enable dllexport or not

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -912,7 +912,9 @@ zsys_udp_recv (SOCKET udpsock, char *peername, int peerlen)
 {
     char buffer [UDP_FRAME_MAX];
     in6addr_t address6;
+#if (!defined (__WINDOWS__))
     inaddr_t *address = (inaddr_t *) &address6;
+#endif
     socklen_t address_len = sizeof (in6addr_t);
     ssize_t size = recvfrom (
         udpsock,
--- a/configure.ac
+++ b/configure.ac
@@ -228,7 +228,7 @@ AC_MSG_CHECKING([whether to install manpages])
 AC_MSG_RESULT([$czmq_install_man])
 
 # Set some default features required by libczmq code.
-CPPFLAGS="-D_REENTRANT -D_THREAD_SAFE $CPPFLAGS"
+CPPFLAGS="-DCZMQ_INTERNAL_BUILD -D_REENTRANT -D_THREAD_SAFE $CPPFLAGS"
 
 # OS-specific tests
 case "${host_os}" in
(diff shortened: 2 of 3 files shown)


sappo:
Hi @diorcety,
thanks for your contribution. Could you please read our contributing guidelines. What I'm missing in your pull request a clear problem statement and the clear explanation of your solution. Is it really necessary the append the CZMQ_EXPORT to all methods in the source files. We do have the exports in the headers which should be enough.

bluca:
Thanks for the fixes!
But the CZMQ_EXPORT is already added to those symbols in the public headers in include/, shouldn't that be enough?

Also, we can merge, but please note that the changes to configure.ac are going to be overwritten, as they are autogenerated from zproject, and specifically from https://github.com/zeromq/zproject/blob/master/zproject_autotools.gsl
If you could please send a PR to add it there too, it would ensure your changes are not lost.

diorcety:
I removed the CZMQ_EXPORT from the c source it was a mistake.

src/zsys.c
```diff
@@ -912,7 +912,9 @@ zsys_udp_recv (SOCKET udpsock, char *peername, int peerlen)
 {
     char buffer [UDP_FRAME_MAX];
     in6addr_t address6;
+#if (!defined (__WINDOWS__))
     inaddr_t *address = (inaddr_t *) &address6;
+#endif
```
sappo: This change seems the solve a different problem!?

diorcety:
The problem is that i'm trying to compile czmq using autotools (./autogen.sh && ./configure && make). libtool trying to compile file twice, one time using -DDLL_EXPORT -fPIC the second time without it. CZMQ_EXPORTS is neved defined using autotools. Maybe there is a propre solution

src/zsys.c
diorcety: Yeah sorry. Unused variable for windows switch.

sappo: Could you please put this into a second commit. You don't need to open another PR. For commit message we do use the Problem/Solution statement, e.g.:

Problem: address variable in zsys is unused in windows build
Solution: only define variable when build os is not windows

sappo:
Okay that's neat :+1: But as @bluca remarked we generate the autotools configuration files with zproject. Do you mind adding your changes to the zproject_autotools.gsl file as well. Then we can just regenerate the file and apply your change to all projects based on zproject e.g. malamute.

diorcety:
Okay i will

diorcety:
So i need a generic term for compilation of library, which is defined only inside

bluca:
Do you mean for the zproject template? If so, you can use the project variable for example:

`$(PROJECT.PREFIX)_INTERNAL_BUILD`

This, at regen time, will be expanded to CZMQ_INTERNALBUILD, MLM_INTERNAL_BUILD etc

sappo:
Use `$(PROJECT.PREFIX)` for the define which accumulates to `$(PROJECT.PREFIX)_INTERNAL_BUILD` and `$(project.name:c)` for the other configure.ac variables.

diorcety:
As you want. I will that then.

diorcety:
https://github.com/zeromq/zproject/pull/626

sappo:
The changes to include/czmq_library.h are generated by zproject as well you can apply them here: https://github.com/zeromq/zproject/blob/master/zproject_class.gsl#L40

diorcety:
https://github.com/zeromq/zproject/pull/627/files

sappo:
@diorcety thanks for your patience and the patches to zproject. It's great that we do have a working mingw build now. Would you mind documenting how to build czmq for mingw64 in the README for other users? That would be nice!

diorcety:
I can if you want but it is really straightforward ./autogen && ./configure && make && make install. Installing the missing part using pacman.

sappo:
It's great that it's straight forward but I wouldn't have known that as a
mingw novice. Thus I think a couple words in the Readme are useful for the
likes of me.
Am 02.06.2016 6:57 nachm. schrieb "Diorcet Yann" <notifications@github.com>:

PR #1458, opened 2016-06-08
Problem: CZMQ out of date with zproject

bluca:
Solution: regenerate


bluca:
@sappo - unfortunately Google announced in the r11 release that they are going to deprecate GCC in the NDK, and that 4.9 is the last version they'll ship, and that folks should starting using Clang instead:

https://developer.android.com/ndk/downloads/revision_history.html (see revision r11 news)

At some point I guess we'll have to have a look at what's needed to switch the Android CI to Clang

sappo:
That's a shame, but I wanted to take a closer look at Clang anyhow. This seems to be an excellent excuse to motivate myself :)

PR #1472, opened 2016-06-30
Problem: [zhashx] packing items will fails if they're chained

sappo:
Solution: Use the total size of the list instead of the prime index to
store intermediate values before they're serialized, which will be
smaller in case items are chained under one prime index. Also introduce a
new value index to catch all values.


bluca:
Bonus points if you add a unit test for the corner case you fixed :-)

sappo:
I need some keys that map to the same bernstein hash. Do you have any idea how to get them? Brute force?

bluca:
No idea beside brute force, sorry!

PR #1473, opened 2016-06-30
zhashx.cpp does not build with MSVC++

joegsn:
VC++ doesn't handle c99 variable length arrays on the stack.


bluca:
Thanks!

For your next contributions, could you please follow the contributing guidelines for the git commit message?

```
Problem: X is broken

Solution: do Y and Z to fix X
```

http://zeromq.org/docs:contributing


Also, if you add something in a commit and then fix it or refactor it in a following one in the same PR, please git rebase and squash them, so that the git history is more readable.

joegsn:
Can do.

PR #1476, opened 2016-07-02
Problem: README is too vague about pkg-config

### LINKED ISSUE #1475: czmq --enable-drafts doesn't completely enable the draft api
I'd been banging my head against a wall for a bit, trying to figure out how to get the client/server socket types working in czmq.  I'd realized yesterday that the functionality required the --enable-drafts parameter to the configure script, but while that seemed to enable them, they didn't actually work.  A very simple program that would create a zsock_new_server, and then assert that it was valid would error out, because the result was null.

It finally occurred to me that zmq also defines a drafts parameter for building itself.  The cflag define is also used in its header.  But maybe the configure script isn't taking that into account.  When I finally ran it with a CFLAGS="-DZMQ_BUILD_DRAFT_API" environment variable set, then it finally built correctly, and I could use the server/client socket types again.

So, in czmq (with libzmq already being configured/compiled/installed with ./configure --enable-draft=yes), the following was needed for me to build with it all enabled.
```env CFLAGS="-DZMQ_BUILD_DRAFT_API" ./configure --enable-drafts ; make ; make install```

I'm minimally knowledgable about autogen scripts, but I believe this can be solved with a simple patch which would make the configure script add that cflag when --enable-drafts is true. (This two-line addition seemed to work for me.)
```
diff --git a/configure.ac b/configure.ac
index e92970a..8f26527 100755
--- a/configure.ac
+++ b/configure.ac
@@ -422,6 +422,8 @@ if test "x$enable_drafts" = "xyes"; then
     AC_MSG_NOTICE([Building stable and legacy API + draft API])
     AC_DEFINE(CZMQ_BUILD_DRAFT_API, 1, [Provide draft classes and methods])
     AC_SUBST(pkg_config_defines, "-DCZMQ_BUILD_DRAFT_API=1")
+    AC_DEFINE(ZMQ_BUILD_DRAFT_API, 1, [Provide draft classes and methods])
+    AC_SUBST(pkg_config_defines, "-DZMQ_BUILD_DRAFT_API=1")
 else
     AC_MSG_NOTICE([Building stable and legacy API (no draft API)])
     AC_SUBST(pkg_config_defines, "")
```

I think there are probably other compile scripts (for maybe cmake, or windows) which might need adjusting as well, to supports the draft api.

bluca:
Solution: clarify what it is used for, how to have the build system
use a different non-standard path for local installs, and how to
achieve the same result manually when it's not available.

Fixes #1475

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,8 @@ Plus some others:
 
 * {{valgrind}} - a useful tool for checking your code.
 
+* {{pkg-config}} - an optional useful tool to make building with dependencies easier.
+
 Which we install like this (using the Debian-style apt-get package manager):
 
     sudo apt-get update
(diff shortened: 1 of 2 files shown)


somdoron:
great!

PR #1490, opened 2016-07-21
Problem: Contribution guidelines are easily missed

sappo:
Solution: Use github pull request templates to make sure new users don't
miss out on reading our contribution guidelines.

(If this works out, we might consider adding those to zproject!)


bluca:
Was about to implement something similar, given all the times I've been copypasting the same comment :-) Thanks!

sappo:
Yes, this might add another step for existing contributers though. Let's see how it works!

bluca:
I'll also add a notice about recursive merge node, since together with the commit message once merged it cannot be fixed anymore.

PR #1493, opened 2016-07-24
Problem: Travis CI OSX builds are broken

bluca:
Solution: add a workaround to reinstall libtool. Travis hasn't fixed
the issue in a month, so time for a little hack until they sort it.


bluca:
Note that since this is (hopefully) temporary I didn't add it to zproject. Should be only a matter of time until Travis fixes it. If they don't then we can port it over.

Issue #1498, opened 2016-07-28
zsys_run_as exits with -1 on Windows even if it's a no-op


ballan91:
zsys_run_as is not yet ported to windows, however should not return -1 if the operation was going to be a no-op due to null import parameters, perhaps a test for no-op scenario could be performed at the top level and we could exit early?

This causes an issue particularly in Malamute on Windows right about here [https://github.com/zeromq/malamute/blob/master/src/malamute.c#L88](url)

bluca:
@sappo - AFAIK on Windows the uid/gid cannot be changed unless the password is provided as well, and obviously we don't want to force users to write the password in plaintext config files.
So I'd argue that zsys_run_as on Windows should enforce that all arguments _are_ NULL with asserts, and return 0, so that programs like Malamute can be sanely cross-platform. What do you think?

sappo:
Sure, if there is no option to make it work on windows it should not crash. The question is though will be program fail anyhow if the command didn't do anything but return successful.

PR #1499, opened 2016-07-28
Problem: zsys_run_as returns -1 for windows even for no-op

ballan91:
Solution: check for no-op and exit early

Hopefully this is ok, I spend my weekdays in Python, it's been a while since I've played with C, though the change is trivial.

This is to address this issue #1498

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -1079,7 +1079,9 @@ zsys_daemonize (const char *workdir)
 
 int
 zsys_run_as (const char *lockfile, const char *group, const char *user)
-{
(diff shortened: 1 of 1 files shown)


sappo:
@ballan91 thanks for your patch! I agree that your problem is valid but the solution is not. We do not return success if the parameters are wrong. Instead we let the application fail with an assert in such cases, e.g.:
```C
assert (lockfile);
assert (group);
assert (user);
```
A valid solution would be to implement the windows part or not to invoke the method with NULL parameters.

sappo:
I'm closing this, because there's no progress foreseeable.

Issue #1501, opened 2016-08-17
zbeacon fails to reconnect when changing IP addresses (Windows, small interval)


twhittock:
zbeacon detects a failure to send, and will then immediately attempt to recreate the socket on whichever interfaces are currently available.

This fails when changing IP addresses, because the bind will assert (address not available) because the interface hasn't finished reconfiguring itself.

Putting a small delay in between the failed send and the reconfigure resolves the issue for me, but it doesn't seem clean. What's the correct method?

```
            if (zsys_udp_send (self->udpsock, self->transmit, &self->broadcast, sizeof (inaddr_t)))
                //  Try to recreate UDP socket on interface
                // <<< HERE WE NEED TO SLEEP OR SOMEHOW DETECT THE INTERFACE IS UP >>>
                s_self_prepare_udp (self);
```

bluca:
Fixed by PR, closing. Thanks!

PR #1503, opened 2016-08-17
Problem: zbeacon won't reconnect on ip address change

twhittock:
Solution: close the socket on bind failure and attempt on the next interval

Fixes issue #1501


bluca:
Will this change make it retry forever if it cannot bind?

twhittock:
Yes - but that's kind of the point of zbeacon... Otherwise when the interface goes down for a long time, zbeacon will never restart. It has no error signaling capabilities, sadly.

PR #1515, opened 2016-09-14
Problem: LIBCZMQ_EXPORTS is still used with some builds such as CMake…

kou:
Solution: Follow CZMQ_EXPORTS -> LIBCZMQ_EXPORTS change

LIBCZMQ_EXPORTS was renamed to CZMQ_EXPORTS at
ef1e42ea9ba481a97629e8dcbd2b36b9f1e3bc7b .

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,3 +59,4 @@ oikosdev
 J. Ritchie Carroll
 Patrik Wenger
 Luca Boccassi
+Kouhei Sutou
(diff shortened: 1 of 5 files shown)


bluca:
Note that those files are all autogenerated by zproject. You'll want to fix zproject templates, otherwise at the next regeneration they will be overwritten.

For example the CMake one is: https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl#L197
And I think it should become "$(PROJECT.PREFIX)_EXPORTS", but you'll need to test it

kou:
I've confirmed the change and sent a pull request: https://github.com/zeromq/zproject/pull/694

bluca:
Thanks!

PR #1517, opened 2016-09-19
Problem: Can't build Release type with "git clone"d source

kou:
Solution: Respect the specified CMAKE_BUILD_TYPE value. If users don't
specify CMAKE_BUILD_TYPE, auto CMAKE_BUILD_TYPE set is used. Normally,
users don't specify CMAKE_BUILD_TYPE, so the behavior isn't changed.


bluca:
@kou note that again this is all generated from zproject so it should be fixed there

kou:
Ah, sorry. I forgot it.
I've sent a pull request: https://github.com/zeromq/zproject/pull/695

PR #1531, opened 2016-10-31
Problem: internal use of deprecated API

bluca:
Solution: see commits

I'll send the removal of v2 deprecated APIs separately.

Note the hack in zsys.c: would be nice to solve it by making the ZMQ v2/v3 socket options available, as they are currently not.
Unfortunately the way it works is that zsock_opt.gsl generates an API file, which does not have the semantics to add IFDEFS per version.
So for now a bit of a hack will do.
Suggestions welcome!

### DIFF
--- a/src/zsock.c
+++ b/src/zsock.c
@@ -531,7 +531,7 @@ zsock_bind (zsock_t *self, const char *format, ...)
     if (zrex_eq (rex, endpoint, "^tcp://.*:(\\d+)$")) {
         assert (zrex_hits (rex) == 2);
         int port = atoi (zrex_hit (rex, 1));
-#if defined (HAVE_LIBSYSTEMD)
+#if defined (HAVE_LIBSYSTEMD) && (ZMQ_VERSION >= ZMQ_MAKE_VERSION (4, 2, 0))
         if (zsys_auto_use_fd ()) {
             int last_handle = SD_LISTEN_FDS_START + sd_listen_fds (0);
             int handle;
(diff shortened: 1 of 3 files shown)


twhittock:
The code was in ZMQ v2 ifdef blocks - isn't this backwards compatibility code?

twhittock:
didn't mean to close that sorry!

bluca:
Yes, but as discussed on the mailing list we are about to retire the v2 APIs so it won't be available anymore :-)
Sending it separately as it will be a big diff, and wanted to draw attention to the zsys hack

twhittock:
sorry I may be being a bit dense here, but isn't the code within the ifdef is only meant to be compiled if compiling CZMQ against a v2 version of libzmq? Since ZMQ_VERSION_MAJOR is currently 4, the #if check would fail, and the other code would be compiled instead?

I guess what I'm saying is that either we remove the whole #if ZMQ_VERSION_MAJOR == 2 block, or keep it the same for compatibility with that version of the library, right?

bluca:
We want to keep it, so that the current version of CZMQ can be used with older versions of libzmq.

The problem is the internal API: zsocket_* is deprecated and about to be removed - when that happens, this change ensures that CZMQ will still work with libzmq 2.x

twhittock:
You're saying there will be retroactive changes to zmq 2.x libraries to remove this function?

bluca:
zsocket_* functions come from here in CZMQ, not from libzmq

twhittock:
Thanks sorry I missed that.

bluca:
No problems, with these many levels I get confused often as well :-)

bluca:
@sappo as discussed on the mailing list I'll send the pr to retire v2 APIs after this is merged

PR #1532, opened 2016-11-01
Problem: time to retire deprecated APIs

bluca:
Solution: remove v2 APIs that have been declared deprecated when
3.0.0 was released: zauth_v2, zbeacon_v2, zctx, zmonitor_v2, zmutex,
zproxy_v2, zsocket, zsockopt, zthread.

Also regenerate to update Windows and bindings.


paddor:
Great! Thank you. I have a feeling the API XML files need to be updated too. I could do this later if needed.

bluca:
I think I got them all, old APIs didn't have a file in api/ but they were mentioned in project.xml from which they are gone, but if I missed something let me know!

paddor:
Ah okay, never mind! My teammate is just having problems. The binding can't find the function `zsock_tos` in the library. But I think that one didn't even get removed. Maybe it's unrelated.

bluca:
Yep that one stayed in, and it's in include/zsock.h - so strange that it can't find it. The definition is inside an ```#if (ZMQ_VERSION_MAJOR == 4)```, which version of libzmq is the czmq library built against?

paddor:
Sorry, false alarm! It was built against 3.2.5! We thought it was master from a few days ago. Thanks anyway!

bluca:
No problem!

PR #1553, opened 2016-11-10
Problem: no NEWS entry for 4.0.1

bluca:
Solution: mention ABI leakage fix, finalize changelog date

As mentioned


sappo:
20s is the best merge time achievable through email latency :-D

bluca:
ahah yes that's a bottleneck, perhaps we could have a zeromq-based push notification to IRC :-D

PR #1555, opened 2016-11-10
Problem: python bindings fail during release build

sappo:
Solution: don't run deploy step on pyhton bindings as they don't
have any.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,4 +84,4 @@ deploy:
   on:
     branch: master
     tags: true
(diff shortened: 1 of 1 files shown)


sappo:
I'm not sure why the bintrayUpload fails though, which isn't an issue because I can do it manually in no time but it blocks the deployment of the czmq-android.jar to github releases :(

bluca:
I've seen that, very annoying, unfortunately I'm not really familiar with that stuff to try and debug it :-/

sappo:
It's my mess to cleanup :) Debugging through travis is just annoying ;)

Issue #1560, opened 2016-11-18
Problem: zsock options could be available at build time but not at runtime


bluca:
I'm working on a solution, here's a brain dump while I think about it.

Distributions have automated dependency generation based on public symbols usage.

So if libfoo has foo () in version 2, and libbar uses it, the libbar package will automatically get a dependency on libfoo >= 2.

But for preprocessor definitions it's more tricky. These are included from headers at build time, so the build system when parsing the shared object has no way to know where they came from. The automatically generated symbols list cover public symbols in shared objects, not defines in headers.

More concretely, I have the case of CZMQ that is built against libzmq 4.2.0, and thus enables the various new socket options. But it doesn't use any of the new public symbols, so the dependency is on libzmq >= 4.1.2:

https://packages.debian.org/stretch/libczmq4

If at runtime I were to try to use, for example, zsock_set_use_fd, and it was done with libzmq 4.1.5, which is legal as far as dependencies are concerned, I would get an assert.

A solution could be to check in the option implementation at runtime, something like:

```
void
zsock_set_heartbeat_ivl (void *self, int heartbeat_ivl)
{
    assert (self);
#   if defined (ZMQ_HEARTBEAT_IVL)
    int major, minor, patch;
    zmq_version (&major, &minor, &patch);
    if (ZMQ_MAKE_VERSION (major, minor, patch) < ZMQ_MAKE_VERSION (4, 0, 0))
        return;
    int rc = zmq_setsockopt (zsock_resolve (self), ZMQ_HEARTBEAT_IVL, &heartbeat_ivl, sizeof (int));
    assert (rc == 0 || zmq_errno () == ETERM);
#   endif
}
```

We do have the means to do it, but we don't track them too well, so they'd have to be fixed first in the XML to be more precise with the version that introduced them.

PR #1561, opened 2016-11-20
Problem: zsock options could be available at build time but not at runtime

bluca:
Solution: rework the sockopts compatibility layer to generate a
runtime check together with the build time check. In case the runtime
check fails, print an error stating what the minimum required is vs
the runtime version. It could be argued that an assert should be used
instead, but this follows the same principle as the existing
implementation where options not available at buildtime are compiled
to simply return 0 and be no-ops, rather than asserting.
Rework the sockopts.xml layout to let the options of each zmq version
be grouped together in different and independent blocks to allow mix
and match. Introduce a major_removed element to indicate in which
version an option has been removed, to allow a finer granularity in
the build time and run time checks.

Fixes https://github.com/zeromq/czmq/issues/1560

The retired options cannot be exposed in the public API unfortunately, as the bindings cannot cope with missing symbols.

### DIFF
--- a/api/zsock_option.api
+++ b/api/zsock_option.api
@@ -8,6 +8,8 @@
 ******************************************************************
 -->
 
+<!-- The following socket options are available in libzmq from version 4.2.0 -->
+
 <method name = "heartbeat ivl" polymorphic = "1">
     Get socket option `heartbeat_ivl`.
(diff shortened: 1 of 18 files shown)


c-rack:
Great 👍

PR #1566, opened 2016-12-05
Problem: zsock.c tests have typos

jimklimov:
Solutions: A couple of fixes to different issues in the tests

Problem2: Some source files are marked executable
Solution: Unmark them


bluca:
Thanks!

Note that a lot of files that had the +x bit are generated from zproject so they might get restored if it's a problem there and it's not fixed: https://github.com/zeromq/zproject

jimklimov:
Yes, I guess they might also be generated, edited or copied from Windows e.g. over samba or natively - which may also botch permissions.

PR #1567, opened 2016-12-05
Problem: zsock.c tests can be unreliable in 32-bit builds

jimklimov:
Solutions: Where I had issues, root out the possible rootcauses, and improve debugging (in case others have similar issues on their platforms)

### DIFF
--- a/src/zhash.c
+++ b/src/zhash.c
@@ -906,6 +906,9 @@ zhash_test (bool verbose)
     srandom ((unsigned) time (NULL));
     for (iteration = 0; iteration < 25000; iteration++) {
         testnbr = randof (testmax);
+        // printf("%d / %d\n", testnbr, testmax);
+        assert(testnbr!=testmax);
+        assert(testnbr<testmax);
         if (testset [testnbr].exists) {
             item = (char *) zhash_lookup (hash, testset [testnbr].name);
             assert (item);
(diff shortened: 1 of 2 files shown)


jimklimov:
FYI: According to https://travis-ci.org/bluca/czmq/jobs/181979236 the `randof()` is still invalid for MacOSX at least with `cmake`:

````
919 13: Test command: /Users/travis/build/bluca/czmq/src/czmq_selftest "--continue" "--verbose" "--test" "zhash"
920 13: Test timeout computed to be: 60

921 13: Assertion failed: (testnbr!=testmax), function zhash_test, file /Users/travis/build/bluca/czmq/src/zhash.c, line 910.

922 13: Running czmq test 'zhash'...
923 13:  * zhash: 
924 13/33 Test #13: zhash ............................***Exception: Other  0.06 sec
925 test 14
926      Start 14: zhashx
927
928 14: Test command: /Users/travis/build/bluca/czmq/src/czmq_selftest "--continue" "--verbose" "--test" "zhashx"
929 14: Test timeout computed to be: 60
930 14/33 Test #14: zhashx ...........................***Exception: Other  0.11 sec
````

Not sure what the failure in case 14 was about, but 13 is that the test number (as an array index) happened equal to testmax (as the zero-based array size), which should not happen (accesses to or beyond `array[testmax]` segfault).

bluca:
Could you please open an issue so we track it? Unless you got a fix ready

jimklimov:
I thought C4 was against opening issues... kind of at all, other than PRs with code ;)
Got no Mac so no way nor know-how for reproduction and fix...

bluca:
it's fine if there's no way to fix it due to platform for example :-)

PR #1568, opened 2016-12-07
Problem: Travis does not install docs build-deps

bluca:
Solution: add build-dependencies and regenerate from zproject to test and deploy in releases all the documentation


.gitignore
```diff
@@ -20,14 +20,18 @@ save.xml
 
 # Executables
 src/czmq_selftest
-src/makecert
```
jimklimov: Strange... did these deliverables disappear, or are they not listed in project.xml?

.gitignore
```diff
@@ -84,12 +99,13 @@ src/app/jni/output
 # Python build directory
 build/
 
-# MSVC build directories
```
jimklimov: I think this block also has to stay, for portable projects at least. Probably `zproject_git.gsl` should take care of this (not a CZMQ issue except further regeneration of the project when that fix lands).

.gitignore
```diff
@@ -84,12 +99,13 @@ src/app/jni/output
...
+# Qt
+*pro.user
+moc_*
+build-*
+bindings/qt/selftest/selftest
 
 # Temporary files
 *.swp
+*.bak
 *~
```
jimklimov: Would it make sense to keep this in common `zproject_git.gsl` as well?

.gitignore
sappo: See https://github.com/zeromq/czmq/commit/d6ffdcc11e7feb997302e4982a95bf140ba058a2

.gitignore
sappo: In general yes, but there are always a couple of exceptions which differ for each project.

.gitignore
bluca: makes sense

.gitignore
jimklimov: Thanks for the link. Seems ok :)

PR #1571, opened 2016-12-08
Problem: Auto-generated documentation files were tracked in git

jimklimov:
Solution: "git rm" the files reported by "make clean" ;)

NOTE: Please let Travis complete all the tests before merging, so we make sure distcheck is not broken by this


bluca:
Note that you can setup a personal traving account by simply logging in with your Github handle:

https://travis-ci.org/

Then you can enable your forks in the account, and it will run on every push to your fork, without needing to open a RP

jimklimov:
Well, the majority of tests passed, and faults were not related to docs. An assertion fail on MacOS an a timeout on Linux, both with a cmake variant.

PR #1576, opened 2016-12-27
Problems: old Travis OSX workaround not removed, libsodium rebuilt everytime

bluca:
Solutions: see commits


bluca:
This was for the czmq-specific ci_build script - next I'll try to implement a general one in zproject

PR #1594, opened 2017-01-08
Problem: zpoller_remove() does not always fail when removing non-existent reader

paddor:
Solution: When using zlist-based implementation (not ZMQ_HAVE_POLLER),
check for existence first. Set errno = EINVAL and return -1 if it
doesn't exist.

### DIFF
--- a/src/zpoller.c
+++ b/src/zpoller.c
@@ -174,8 +174,14 @@ zpoller_remove (zpoller_t *self, void *reader)
     else
         rc = zmq_poller_remove_fd (self->zmq_poller, *(SOCKET *) reader);
 #else
-    zlist_remove (self->reader_list, reader);
-    self->need_rebuild = true;
+    if (zlist_exists (self-reader_list, reader)) {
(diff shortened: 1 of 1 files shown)


src/zpoller.c
```diff
@@ -174,8 +174,14 @@ zpoller_remove (zpoller_t *self, void *reader)
     else
         rc = zmq_poller_remove_fd (self->zmq_poller, *(SOCKET *) reader);
 #else
-    zlist_remove (self->reader_list, reader);
-    self->need_rebuild = true;
+    if (zlist_exists (self-reader_list, reader)) {
```
bluca: @paddor typo here, missing ">"

bluca:
Good catch, had the same problem with the lack of feedback from zlist_remove in zmsg.

src/zpoller.c
paddor: Oh geez. Thanks!

paddor:
@bluca Just looked that up. Yours is a more performant solution. I'll file another PR.

PR #1597, opened 2017-01-08
Problems: helper functions symbols leaked, zmsg_gossip private class symbols leaked when building with DRAFT enabled, no automated testing of private classes

bluca:
Solutions: see commits

Note that I did not commit the regenerated doc/Makefile.am as there is a problem ATM, see https://github.com/zeromq/zproject/issues/806#issuecomment-271090868

The symbols leaks only affects builds with DRAFT on so it's not a big issue, no need for a release to fix it


jimklimov:
See PR#821 regarding the doc/Makefile, hopefully that should fix it.

Issue #553, opened 2014-07-14
czmq crash on exit on Windows


twhittock:
czmq relies on an atexit handler to close out sockets. This crashes on Windows with a spurious error message. It complains that Winsock has not been initialised, even though it has been initialised (and has not been cleaned up.) I debugged through and ensured that there were fewer calls to WSACleanup than WSAInitialise.

It appears that any winsock operations in an atexit handler do not work? Although I'm unable to find any documentation on this topic.

Removing the atexit() call and hacking the uninitialise method to change its visibility, so I can call it at the end of main() makes the issue go away for me.

PR #1600, opened 2017-01-10
allow clean exit of executable on windows

bbdb68:
Problem : czmq tests fail on windows

Solution : add zsys_shutdonw() before exit

this is an experiment, let's discuss this before merge

### DIFF
--- a/src/czmq_selftest.c
+++ b/src/czmq_selftest.c
@@ -202,6 +202,7 @@ main (int argc, char **argv)
     else
         test_runall (verbose);
 
+    zsys_shutdown();
     return 0;
 }
 /*


bbdb68:
adding the shutdown helped me in my projects, but here it seems to be a failure :-1:

bluca:
So zsys_shutdown should be called automatically as it's registered as an at_exit callback

bluca:
See: https://github.com/zeromq/czmq/blob/master/src/zsys.c#L187
I think that should work on Windows, right?

From the logs on Appveyor it looks like all the individual tests are failing, or rather not even running given the runtime. It's really weird :-(

But thanks for looking into this! It's been broken forever

bluca:
This doc suggests it should work on Windows too: https://msdn.microsoft.com/en-us/library/tze57ck3(v=vs.140).aspx

bbdb68:
@bluca from what I see, I makes a real difference in the tests when I add the zsys_shutdown call (as I did above). Maybe the point is that I compile czmq as a dll ?

bbdb68:
form the doc you showed me :+1: 
"The code in the atexit function should not contain any dependency on any DLL which could have already been unloaded when the atexit function is called."

In my case ,the atexit function is _in_ the dll, so it could be unloaded before atexit run...

bluca:
Ah, I see.

Could you please wrap it in an ifdef so it doesn't affect other systems?

```
#if defined (__WINDOWS__)
zsys_shutdown
#endif
```

bluca:
Also, do you have any idea if it's possible to have a similar functionality for a DLL? The library relies on that to do a clean shutdown, so it would be helpful if it worked in that case too

twhittock:
Please see also #553

bluca:
Not very promising :-(
Is there a way to detect at build time if it's going to be a static or DLL build?

If so, I'd propose to skip the atexit() registration when building as DLL on Windows, and adding to the documentation that a manual zsys_shutdown() is necessary in those cases

bbdb68:
@twhittock  thanks for the heads up
I wonder why there is not such issue with the zmq library, since it uses winsock as well ?

bluca:
libzmq is a low-level library and does not implement automated setup and teardown like CZMQ does

bbdb68:
to get back to the intial subject (test crash on appVeyor), is there a way to browse the build directory on appveyor ? this could help to diagnose the issue ? or can the appveyor script be run on a local machine ?

I notice that the czmq (and zyre) cmakeList builds there executables (selftest for example)  in the _source_ directory, not in the build directory where the dlls are built, so on windows with the provided CMakeLists, you cannot run the executables "as is". You must either copy the executables to the directory where the dlls are build, or the reverse. BTW this makes debugging of these projects on windows really painful.

If the same happens on appVeyor, the self_test.exe just cannot run, without a copy step.
Looking at the appVeyor.yml file, I see copy of the libsodium and libzmq files in "after build", but no copy of the czmq dll it self (provided ctest runs in the selfttest_exe directory ?)
that's why browsing the build dir would be interesting...

bluca:
I haven't looked too deeply into Appveyor but I don't think it's possible to browse the directories.

As a workaround, you could add some commands to the script to print the directory listings manually.

Regarding CMake directory, do you think you could fix it? The CMake files are generated by:

https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl

But if you fix it here first, then we can port it there if it works.

bbdb68:
for the camke issue, my guess is that the issue is there
https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl#L304
I would try to remove the set_target_properties command, but how can this be tested ?

bluca:
You can run appveyor on your forks, it's very easy.

Go to https://www.appveyor.com/ and login with your Github account, and then enable your czmq fork. From that point every time you push it will run the tests.

I did that but it seems to make it worst:

```
7z a -y -bd -mx=9 czmq.zip "%APPVEYOR_BUILD_FOLDER%\src\%CONFIGURATION%\zmakecert.exe" "%APPVEYOR_BUILD_FOLDER%\src\%CONFIGURATION%\czmq_selftest.exe" czmq.dll libsodium.dll libzmq-*.dll
7-Zip [64] 16.04 : Copyright (c) 1999-2016 Igor Pavlov : 2016-10-04
Scanning the drive:
WARNING: The system cannot find the file specified.
C:\projects\czmq\src\Debug\czmq_selftest.exe
4 files, 3787776 bytes (3699 KiB)
Creating archive: czmq.zip
Items to compress: 4
Files read from disk: 4
Archive size: 985065 bytes (962 KiB)
Scan WARNINGS for files and folders:
C:\projects\czmq\src\Debug\czmq_selftest.exe : The system cannot find the file specified.
----------------
Scan WARNINGS: 1
```

https://ci.appveyor.com/project/bluca/czmq/build/job/93ni4j5rolt51m2a

bbdb68:
Hi

 regarding build/cmake issue I will try to spend some time on it, not sure of the outcome

regarding the exit issue and the design of czmq (automatic setup/teardown) it sounds diffcult to have it cleanly working in a windows dll, though I am not a windows compilation expert... and it could lead to much wider discussion... there are so much topic about static variables in windows dll everywhere...

(example : I have the exit issue on czmq_selftest, but not on zyre/zpinger, now that close event is handled)

So I guess this PR could be closed, and if I have some result on cmake I will open a new one

Bruno

bluca:
Ok, thank you!

PR #1601, opened 2017-01-12
Fix win32 build

bbdb68:
problem : unit tests do not run on appveyor

solution :
the issue is that the exe and dlls were not on the same folder, so I propose to build czmq_selftest and zmakecert in the same folder as the dlls. BTW, this makes debugging wihtin visual studio much easier.
The explicit call to zsys_shutdown is also required.

Please note that the solution propose here implies the change in a generated file (CMakeLists.txt) so the real fix should be in zproject I guess, so this may not be merged "as is"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,10 +299,10 @@ target_link_libraries(
     ${LIBZMQ_LIBRARIES}
     ${OPTIONAL_LIBRARIES}
 )
-set_target_properties(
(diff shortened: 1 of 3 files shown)


bbdb68:
see also 
https://github.com/zeromq/zproject/pull/833

bluca:
Fantastic, thank you! 

I missed the change in the appveyor file when I tried this.

PR #1602, opened 2017-01-12
ZSYS_INTERFACE variable can be a single digit

bbdb68:
problem : 
on windows system, in exotic languages, the interfaces names can be complex, include spaces and special characters, making the use of the variable ZSYS_INTERFACE for network interface selection difficult.

solution:
allow the ZSYS_INTERFACE variable to be a single digit, selecting the i-th number interface of the system.
keep the existing behaviour for other values. 

can be tested/validated with zbeacon self test

### DIFF
--- a/src/zbeacon.c
+++ b/src/zbeacon.c
@@ -101,6 +101,32 @@ s_self_prepare_udp (self_t *self)
         send_to = INADDR_BROADCAST;
         found_iface = 1;
     }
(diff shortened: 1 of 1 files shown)


sappo:
@bbdb68 for your next contribution it would be great if the problem and solution explanation is part of the commit message instead of the pull request. As the commit messages are part of the git history.

Issue #1606, opened 2017-01-15
on windows/debug, zproxy test fails


bbdb68:
here it output

Running czmq test 'zproxy'...
 * zproxy:
I: 17-01-15 19:09:09 zproxy: API command=FRONTEND
I: 17-01-15 19:09:09 zproxy: - FRONTEND type=PULL attach=inproc://frontend authe
ntication=NONE
I: 17-01-15 19:09:09 zproxy: API command=BACKEND
I: 17-01-15 19:09:09 zproxy: - BACKEND type=PUSH attach=inproc://backend authent
ication=NONE
I: 17-01-15 19:09:09 zproxy: API command=PAUSE
I: 17-01-15 19:09:09 zproxy: API command=RESUME
I: 17-01-15 19:09:09 zproxy: API command=CAPTURE
I: 17-01-15 19:09:09 zproxy: API command=$TERM
Assertion failed: sink, file C:\dev\ZeroMQ\github\czmq\src\zproxy.c, line 551

PR #1607, opened 2017-01-16
problem : zproxy self test fails under windows

bbdb68:
solution : replace ipc with inproc (only for windows)


bbdb68:
solves #1606

PR #1608, opened 2017-01-19
Problem: atexit does not work when building a Windows DLL

bluca:
Solution: put more emphasis in the docs, and call zsys_shutdown from each selftest when building on Windows


bluca:
@sappo ping

sappo:
Sorry

bluca:
Thanks! No worries

PR #1609, opened 2017-01-21
problem : zdir test fails on windows/release

bbdb68:
fix : rework the assert macros to avoid calls within assert.


bluca:
Wow all tests passing, finally! Thank you for all your work @bbdb68 !

sappo:
Neat!

PR #1619, opened 2017-02-07
Problem: windows build via build.bat does not have enable/disable draft api switch

rodrigomadruga:
Solution: expose the optional configuration using a property in `czmq.import.props`. Add the option to VS UI in `libczmq.import.xml`. Use the condition in `libczmq.vcxproj` to include draft classes only when option is true.

To use the switch, pass `/p:Option-draftapi=[false|true]` to msbuild. Example from `build.bat`:

`msbuild /m /v:n /p:Configuration=DynDebug /p:Platform=Win32 /p:Option-draftapi=false %packages% %solution% %target% >> %log%`

Probably these changes should go to zproject in someway, as the involved files have the "THIS FILE IS 100% GENERATED BY ZPROJECT; DO NOT EDIT EXCEPT EXPERIMENTALLY" comment.

### DIFF
--- a/builds/msvc/vs2010/build.bat
+++ b/builds/msvc/vs2010/build.bat
@@ -23,8 +23,8 @@ IF NOT EXIST %environment% SET environment="%programfiles%\%tools%"
 IF NOT EXIST %environment% GOTO no_tools
 
 SET packages=
-IF EXIST "..\..\..\..\libzmq\builds/msvc/vs2010\libzmq.import.props" (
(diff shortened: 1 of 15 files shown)


rodrigomadruga:
Right now, there is no way to automatically use the enable/disable draft selection made by` configure.bat` in `build.bat`.

Help in which would be the best way to expose/check this option would be welcome...

bluca:
Thanks!

Note that these changes will be overwritten by the next zproject regeneration. Could you please apply them to https://github.com/zeromq/zproject/blob/master/zproject_vs20xx.gsl ?

PR #1624, opened 2017-02-15
Problem: Wikidot docs are not automatically updated

sappo:
Solution: Add a new build script that will upload the master
revision of the docs to Wikidot on every Non-PR build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,9 @@ env:
   global:
     # Note: Secure variables must be global!
     #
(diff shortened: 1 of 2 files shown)


bluca:
great stuff, thanks!

sappo:
It does not work for deployments though! Why? Because PERL :(

bluca:
lol!
what's the issue?

sappo:
The versions for which the guide is generated is manually set in bash and perl.

PR #1633, opened 2017-03-06
Problem: Adding/Removing network adapters in Windows can cause a null…

kaidoe:
… pointer exception

Solution: According to MSDN GetAdapterAddresses may need to be called multiple times in cases where a device has been added/removed.

### DIFF
--- a/src/ziflist.c
+++ b/src/ziflist.c
@@ -275,9 +275,28 @@ s_reload (ziflist_t *self, bool ipv6)
     DWORD rc = GetAdaptersAddresses (AF_INET, GAA_FLAG_INCLUDE_PREFIX, NULL, NULL, &addr_size);
     assert (rc == ERROR_BUFFER_OVERFLOW);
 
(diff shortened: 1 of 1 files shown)


src/ziflist.c
```diff
@@ -275,9 +275,28 @@ s_reload (ziflist_t *self, bool ipv6)
     DWORD rc = GetAdaptersAddresses (AF_INET, GAA_FLAG_INCLUDE_PREFIX, NULL, NULL, &addr_size);
     assert (rc == ERROR_BUFFER_OVERFLOW);
 
-    PIP_ADAPTER_ADDRESSES pip_addresses = (PIP_ADAPTER_ADDRESSES) zmalloc (addr_size);
-    rc = GetAdaptersAddresses (AF_INET,
-                               GAA_FLAG_INCLUDE_PREFIX, NULL, pip_addresses, &addr_size);
+    static const int maxTries = 3;
```
bluca: This means it will do only 3 tries for the lifetime of the process. Is this what you intended?

PR #1636, opened 2017-03-13
Problem: can't deal with subprocesses via czmq

vyskocilm:
Solution: add a zsubproc class. Unix pipes on steroids. Standard input and ouptputs MAY be connected to zeromq sockets and zsubproc will forward data between pipes and zeromq sockets for you.

Note: I do not like API that much especially zsubproc_run looks like execve and there may be nicer API to use. Plus some methods have weird semantics, or asserts called twice. Marked as draft to have a time to polish the interface.


include/zsubproc.h
```diff
@@ -0,0 +1,112 @@
...
+CZMQ_EXPORT int
+    zsubproc_returncode (zsubproc_t *self);
+
+//  subprocess pid, asserts if process was not started
+CZMQ_EXPORT int
+    zsubproc_pid (zsubproc_t *self);
+
+//  pid of a subprocess, 0 if not yet started
+CZMQ_EXPORT int
+    zsubproc_pid (zsubproc_t *self);
```
sappo: defined twice !?

src/zsys.c
```diff
@@ -251,7 +251,8 @@ zsys_shutdown (void)
     s_sockref_t *sockref = (s_sockref_t *) zlist_pop (s_sockref_list);
     while (sockref) {
         assert (sockref->filename);
-        zsys_error ("dangling '%s' socket created at %s:%d",
+        zsys_error ("[%d]dangling '%s' socket created at %s:%d",
+                    getpid (),
```
bluca: Isn't this POSIX-specific? Windows builds won't be happy I think

bluca:
Nice! But Windows builds don't look happy:

https://ci.appveyor.com/project/zeromq/czmq/build/job/9sg370d9ch0ej861

You might need to ifdef out the implementation and add stubs when building on windows

sappo:
This looks neat! Can you give a an example usage?

Regarding the name zsubproc feels unwieldy maybe just zproc or zpipe

bluca:
There's already a zproc :-)

vyskocilm:
I'd prefer zproc, but there is draft CLASS zproc already - on the other hand this is just a bunch od static methods to manipulate with a process state, so can be moved to zsys (zsys_proc_foo) or we can merge those two classes. I was afraid it'd be confusing.

vyskocilm:
@sappo yes, I plan to extend unit test. Does it count as a documentation? :)

sappo:
Absolutely, maybe have a look at ztrie_test. I tried to make the test cases
as user friendly as possible.

Am 13.03.2017 1:43 nachm. schrieb "Michal Vyskocil" <
notifications@github.com>:

PR #1638, opened 2017-03-14
Problem: Travis CI fails

vyskocilm:
Solution: fix make distcheck, binary is searched in _build/../

@sappo @bluca @jimklimov - maybe we should fix this on zproject level. For instance export some special variable pointing to source directory for distcheck or so.

So the code will be way more simple. Of course FOO is a stupid name, it should be ZFOO or ZSUSHI at least ;-)

    #include <czmq.h>
    int main () {

        // done in selftest binary
        assert (getenv ("FOO"));

        char *zsp = zsys_sprintf ("%ssrc/zsp", getenv ("FOO"));
        zsys_info ("zsp=%s", zsp);
        zstr_free (&zsp);
    }

### DIFF
--- a/include/zsubproc.h
+++ b/include/zsubproc.h
@@ -20,6 +20,9 @@ extern "C" {
 
 //  @interface
 //  Create a new zsubproc
+//  Current version works well on Linux and OSX with libzmq4.
+//  On Windows and with libzmq3 and libzmq2 this function does
+//  return NULL
 CZMQ_EXPORT zsubproc_t *
     zsubproc_new (void);
 
--- a/src/zsubproc.c
+++ b/src/zsubproc.c
@@ -693,7 +693,11 @@ zsubproc_test (bool verbose)
     char *const xargv[] = {"zsp", "--stdout", NULL};
     char *const xenvp[] = {"PATH=/bin/:/sbin/:/usr/bin/:/usr/sbin", NULL};
 
-    zsubproc_run (self, "src/zsp", xargv, xenvp);
+    char *file = "src/zsp";
+    if (zsys_file_exists ("_build/../src/zsp"))
+        file = "_build/../src/zsp";
+
+    zsubproc_run (self, file, xargv, xenvp);
 
     zpoller_t *poller = zpoller_new (zsubproc_actor (self), zsubproc_stdout (self), NULL);
 
--- a/src/zsubproc.c
+++ b/src/zsubproc.c
@@ -139,6 +139,10 @@ struct _zsubproc_t {
 zsubproc_t*
 zsubproc_new ()
 {
+#if defined (__WINDOWS__)
+    zsys_error ("zsubproc_set_stdin not implemented for Windows");
+    return NULL;
+#else
 #if ZMQ_VERSION_MAJOR < 4
     zsys_error ("Cannot use zsubproc with zmq older than 4");
     return NULL;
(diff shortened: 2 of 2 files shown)


vyskocilm:
AppVeyor said  Test zdir failed, however I did not touch this code. zsubproc builds and run w/o problem.

bluca:
Zdir on Windows sometimes fails, don't worry about it.

Some tests are failing on Linux still though, would you like to wait to merge until they pass?

jimklimov:
Look at selftest support PRs we did recently, they may (be used to) handle this case as well.

vyskocilm:
Hi,

I would like to make that work at least against libzmq 4.

For older versions, I would simply torn it off, so users of older zmq would
not be able to use this class.

Sounds good?

Dne 14. 3. 2017 12:30 napsal uživatel "Jim Klimov" <notifications@github.com

bluca:
Sure that's fine. Please document it, and also have both a build time and a runtime check for the version as it's done for the socket options:

https://github.com/zeromq/czmq/blob/master/src/zsock_option.inc

Out of curiosity, what is required from newer versions exactly?

vyskocilm:
Not sure, it uses inproc exclpair sockets to communicate. It simply stops
to work under older versions.

Someone needs to take a look on whats happening here.

Dne 14. 3. 2017 1:44 PM napsal uživatel "Luca Boccassi" <
notifications@github.com>:

jimklimov:
Regarding "improved selftest support" mentioned above, see https://github.com/zeromq/zproject/pull/871

The general idea is that if your self-test routine requires read-only data (fixtures stored in SCM) or data files/dirs generated by test itself (with RW access) you regenerate with recent zproject and the scaffolding guarantees that you have dirs predefined for that (RO data copied to under builddir; RW data just living and dying under builddir), gitignored and distcleaned properly. In the test routine define paths with this template and use them where you name files for this or that purpose:

````
    // Note: If your selftest reads SCMed fixture data, please keep it in  
    // src/selftest-ro; if your test creates filesystem objects, please  
    // do so under src/selftest-rw. They are defined below along with a  
    // usecase (asert) to make compilers happy.  
    const char *SELFTEST_DIR_RO = "src/selftest-ro";  
    const char *SELFTEST_DIR_RW = "src/selftest-rw";  
    assert (SELFTEST_DIR_RO);  
    assert (SELFTEST_DIR_RW);  
    // Uncomment these to use C++ strings in C++ selftest code:  
    //std::string str_SELFTEST_DIR_RO = std::string(SELFTEST_DIR_RO);  
    //std::string str_SELFTEST_DIR_RW = std::string(SELFTEST_DIR_RW);  
    //assert ( (str_SELFTEST_DIR_RO != "") );  
    //assert ( (str_SELFTEST_DIR_RW != "") );  
    // NOTE that for "char*" context you need (str_SELFTEST_DIR_RO + "/myfilename").c_str()  
````

**Comment by bluca** [comment]
It's mostly good now, so I'll merge to avoid letting it open for too long which is not nice

**Comment by bluca** [comment]
There's 2 problems still to solve:

1) Android build fails:

```
src/zsp.c: In function 'main':
src/zsp.c:85:13: error: implicit declaration of function 'getline' [-Werror=implicit-function-declaration]
             ssize_t r = getline (&message, &size, stdin);
```

Likely some arm-specific shenanigans

2) valgrind is very, very angry:

https://travis-ci.org/zeromq/czmq/jobs/211252589

Most likely some cleanup in the subprocess is failing

sappo:
Oh no my android build :-(
After a quick google search replacing `getline` with `fgets` should solve the problem. `getline` seems to have a string glibc dependency.

bluca:
fgets sounds like a good choice

PR #1640, opened 2017-03-16
Problem: valgrind build fails

vyskocilm:
Solution: destroy the poller and add a suppression for glibc thread pool. Plus document the test and code better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,8 +58,9 @@ matrix:
       apt:
         packages:
         - openjdk-7-jdk
-  - env: BUILD_TYPE=docs
-    os: linux
+# temporary disabled, does not work
+#  - env: BUILD_TYPE=docs
+#    os: linux
 
 addons:
   apt:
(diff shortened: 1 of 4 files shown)


vyskocilm:
Local memcheck pass, but not on Travis. I'll check it why.

@bluca @sappo - any idea about docs test, it looks strange to me.

vyskocilm:
OK, memcheck looks like the problem with execve and the fact binary was not found or executed properly. I skipp the test in this case, need to make this resource allocation easier to use from tests.

sappo:
@vyskocilm I started the work on docs but haven't gotten around to finish it. Thanks for deactivating them.

vyskocilm:
So, can be merged.
 * The getline is over, zsp use zfile instead. Fixes android
 * Valgrind test is skipped on Travis as I can't find zsp binary there.
 * docs target is disabled
 * appveyor fails on zdir only

bluca:
Thanks!

PR #1644, opened 2017-03-20
Problem: Windows only sends on one interfaces with INADDR_BROADCAST set (interface option "*")

kaidoe:
Sending to INADDR_BROADCAST on Windows only results in a single packet being sent over the interface with the highest metric.

This results in nodes not being seen on all interfaces when using the any ("*") interface, metric can be observed using the command 'route print'

Solution: SIO_GET_INTERFACE_LIST offers a cheap way of enumerating interfaces available to a specific socket so a packet can be sent per interface. Profiling shows this method at 0.006ms per invokation against 0.5ms for GetIpAddrTable meaning that this can be used on every beacon send instead of caching the results.

### DIFF
--- a/src/zbeacon.c
+++ b/src/zbeacon.c
@@ -417,6 +417,52 @@ s_self_handle_udp (self_t *self)
 
 
 //  --------------------------------------------------------------------------
+//  Send the beacon over UDP
+
+int
+s_emit_beacon (self_t *self)
+{
+#if defined (__WINDOWS__)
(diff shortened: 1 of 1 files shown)


bluca:
Thanks!

If you have the time, would be great to get the IPv6 version working for Windows :-)

kaidoe:
I'll look into it, and thanks for cleaning up the commit.

bluca:
No worries!

PR #1648, opened 2017-03-24
Problem: zproc_run method can't be expressed as zproject API model

vyskocilm:
Solution: see pull requests

In fact code is way more straightforward than before, because argv/env are built right before execve and not passed from zproc_run to actor and pipe handler.

### DIFF
--- a/api/zproc.api
+++ b/api/zproc.api
@@ -20,6 +20,12 @@
         Destroy zproc, wait until process ends.
     </destructor>
 
+    <method name = "set args" >
+        Setup the command line arguments, the first item must be an (absolute) filename
+        to run.
+        <argument name = "args" type = "zlistx" />
+    </method>
+
     <method name = "set stdin" >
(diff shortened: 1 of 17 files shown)


bluca:
Nice! But:

```
src/zproc.c: In function 'zproc_test':
src/zproc.c:986:17: error: unused variable 'xenvp' [-Werror=unused-variable]
src/zproc.c:985:17: error: unused variable 'xargv' [-Werror=unused-variable]
src/zproc.c: At top level:
src/zproc.c:173:1: error: 'arr_size' defined but not used [-Werror=unused-function]
```

want to have a look before or after merging?

vyskocilm:
How it's possible make run well locally? I will fix it

Dne 24. 3. 2017 2:26 PM napsal uživatel "Luca Boccassi" <
notifications@github.com>:

bluca:
Different compiler version/setting? Those are 2 warnings treated as errors

bluca:
Thanks!

PR #1649, opened 2017-03-24
problem :

bbdb68:
czmq cannot be restarted

solution:

ensure static pointers are reset to NULL within zsys_shutdown


bluca:
Thanks!
Perhaps we should add this use case to the zsys_test, given you are not the first who wants to use it

PR #1651, opened 2017-03-26
problem : pull request #1649 did not contain test

bbdb68:
* pull request #1649 did not contain test

solution :

* add shutdown /init sequence in zsys test


bluca:
Thanks!

bluca:
For future commits, please keep the "Problem" statement on one line, so that it shows up in the git shortlog (eg: "Problem: foo bar is broken")

PR #1657, opened 2017-03-31
Off-by-one error on expansion of base64 + null termination

donbowman:
Problem: s_base64_decode() has an off-by-one error on expansion. This shows up under valgrind as a corruption.


donbowman:
i made a mistake in the tests, sorry, will correct

Issue #1659, opened 2017-04-07
Problem: zsys_init() + fork() causes unxpected behavior


jpitc:
### Background

The application I'm working on explicitly requires process isolation, so I'm not using threads. The child processes never `exec()`, also for good reason. I'm working with the git for upcoming CZMQ 4.0.3.

### Problem
Very early, the parent creates a logging socket for the child processes to connect to after it is  spawned. I've been experiencing all manner of strange behavior by the child processes with this pattern.

### Things I've Tried
- The child calls `zsock_destroy()` on all the parents sockets when it starts (just as it would close pipe fd's), this is the proper thing to do but doesn't solve the problem.
- I've tried calling `zsys_shutdown()` first thing in the child to wipe the slate clean, but any subsequent call to czmq triggers `zsys_init()` then fails with `assert(!s_process_ctx)`.
- I've NULL'ed the pointer after its freed and got a malloc SEGV due to a double free... and so on.

### Diagnosis
 
`zsys.c` constains a non-trivial amount of mutable global state. When almost any czmq call is made, that global state is modified, and `zsys_shutdown()` does not restore it. a subsequent call to `zsys_init()` then encounters an undefined initial state and bad things happen.

### Proposed solution

I've added a new internal (static) `zsys_reset()` function which restores all global state to initial values, which `zsys_shutdown()` calls it as its (nearly) final act. This change has reNew contributorsolved all the issues I've seen. So far.

###  Points For Discussion

- CLASS destructors NULL references after they are freed, but `zsys_shutdown()` doesn't follow this
pattern for `zsys.c` global state. Possibly this needs to be cleaned up in the entire tree. Haven't looked.
- My solution duplicates the static initialization of global state into the `zsys_reset()` code, which is a  brittle. If someone in the future adds a some more global state and forgets to update `zsys_reset()`...
- There may be other global state outside zsys, which would complicate matter. Guidance from principles welcome.

bbdb68:
does your code include this fix

https://github.com/zeromq/czmq/pull/1649

It may help

jpitc:
Thanks, yes, my #1660 can be considers a superset of #1649.

sappo:
Hi @jpitc, 
I'm still trying to understand your exact problem. When do you fork? Before or after initializing the zmq context (implizit by first use of czmq socket API)?

If you do it before then this is currently considered an anti-pattern as the context mustn't be shared by multiple processes.

jpitc:
[...]

PR #1660, opened 2017-04-07
Problem: zsys_init() + fork() causes undefined behaviour

jpitc:
closes https://github.com/zeromq/czmq/issues/1659.

First PR here. I've read and agree to C4.1.

This change requires (some) version change, because code that depends on this fix must have a way to check for it. Not sure how new minor/patch versions are handled in the project.

Follow up to #1649 (as it turns out), extending it to reset all global state, not just dangling pointers.

### DIFF
--- a/include/zsys.h
+++ b/include/zsys.h
@@ -36,7 +36,8 @@ CZMQ_EXPORT void *
 //  when the process exits; however this call lets you force a shutdown
 //  earlier, avoiding any potential problems with atexit() ordering, especially
 //  with Windows DLL builds, where atexit() does not work and zsys_shutdown has
-//  to be called manually.
+//  to be called manually. Resets all zsys global state (i.e., HWM, LINGER, ...)
+//  to their initial values.
 CZMQ_EXPORT void
     zsys_shutdown (void);
 
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -219,6 +219,44 @@ zsys_init (void)
     return s_process_ctx;
 }
 
+//  --------------------------------------------------------------------------
+//  Restores all CZMQ global state to initial values. Internal function called
+//  by zsys_shutdown().
+
+static void
+zsys_reset (void){
+    s_process_ctx = NULL;
+    zsys_interrupted = 0;
+    zctx_interrupted = 0;
+
+    s_first_time = true;
+    handle_signals = true;
+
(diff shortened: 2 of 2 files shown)


bluca:
Thanks.

I understand wanting to reset the global state, but there are 2 issues:

1) uncoditionally clearing the context variable means leaking when the shutdown was not done properly due to errors in the application code - it's better to leave it as-is and cause an assert, so that it can be looked at
2) this causes a backward incompatible change in the side effects of a public API, so we can't just release it, it will have to be behind the DRAFT flag and released only the next time there's an API breakage
alternatively, a new public API can be added, eg: zsys_shutdown_reset - if you need that, feel free to send a followup PR

jpitc:
Thanks for explaining. It sounds like you didn't mean to merge this, but you did. Slip of the mouse?

bluca:
It was intentional - as you have read in the C4, we always merge PRs, apart from a few corner cases where things are really, really, really broken. This is not one of them.
I'll send a followup PR with the improvements I mentioned in a few minutes

jpitc:
That's gonna take some getting use to.

jpitc:
In that case, there's no reason not to call `zsys_reset()` in the case of a successful shutdown, right?
There's no contract that promises dirty state after a proper shutdown and restart.

bluca:
Release engineering of a public library is a delicate problem - even if there's no explicit promise, de-facto the current behaviour of these public APIs is that a shutdown follow by a re-init does not discard the previous settings of those various global context variables.
Whether we like it or not, this is how this API currently works, and there might be users relying on this, intentionally or not. For example, one might be enabling IPv6 at the beginning of an application, and then fork. Changing it would remove IPv6 from the child process.

This means this change can't go live (without the DRAFT flag) until we are ready to do an API-breaking release, which is very rare and should be at most once every couple of years.
If you'd like this change to go live sooner, as I mentioned you can add a new API call, for example called zsys_shutdown_reset, which combines both. This removes the risk of affecting existing users.

As a famous developer once said, if people are using it, it's not a bug, it's a feature.

jpitc:
https://github.com/zeromq/czmq/pull/1662#issuecomment-292512244

bluca:
Yes that's true, I hadn't considered that it was broken before, only that it was available

jpitc:
:+1:

PR #1662, opened 2017-04-07
Problem: Global zsys state not cleared after clean zsys_shutdown()

jpitc:
Solution: reset state if shutdown was clean. Until #1649, no code could rely on behavior wrt global state on a zsys shutdown+init (restart), because any attempt to call zsys_init() after zsys_shutdown() would assert and #1649 has not been part of of any official release yet , so there's no reason to punt on this for reasons of back-compat.

followup to  #1649 , #1660 , #1661

### DIFF
--- a/include/zsys.h
+++ b/include/zsys.h
@@ -36,7 +36,8 @@ CZMQ_EXPORT void *
 //  when the process exits; however this call lets you force a shutdown
 //  earlier, avoiding any potential problems with atexit() ordering, especially
 //  with Windows DLL builds, where atexit() does not work and zsys_shutdown has
-//  to be called manually.
+//  to be called manually. A succesful shutdown resets the zsys global state
+//  (HWM, LINGER, etc).
 CZMQ_EXPORT void
     zsys_shutdown (void);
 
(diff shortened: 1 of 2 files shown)


bluca:
Please see https://github.com/zeromq/czmq/pull/1660#issuecomment-292496860

The correct way to have this change published sooner is to add a new API, eg: zsys_shutdown_reset, that does both

jpitc:
@bluca, I've read that carefully (thanks for explaining carefully). I have every respect for back-compat, and in other projects have acted the curmudgeon against other developers on that point. But I think this vigilance is simply misplaced here. To reiterate, the behavior this changes has never been exposed in an official release and is not documented (either way) anywhere. It cannot qualify as a a breaking change by any reasonable standard. 

Even the mere question of what the behavior *is* was a moot point until two weeks ago when #1649 was merged, because anyone who tried to exercise it was stopped by an assert before they got there.

Seeing as you merged my PR and then quickly changed it back, it's clear that "it's in git between version" isn't a de-facto contract to users. To be clear, this only affects a clean shutdown. I took your point about leaking sockets treated as a halting condition.  

So please reconsider whether you're actually protecting users (who ship products against a git version) or throwing a way an opportunity to correct an oversight that will then become written in stone for a year or two. 

But if you still refuse to reconsider, I'll drop the issue....

... And consider you wrong wrong wrong till my dying day :wink:

jpitc:
Answered there.

PR #1663, opened 2017-04-08
Problem: dangling pointers are bug futures

jpitc:
Solution: unify free/nullify into a FREE_AND_NULL macro

whenever a pointer is free()'d but not nullifies, a double-free or user-after-free 
bug is at least potentially possible.
For example #1649 is in part a spot fix for this problem.  Everywhere in
zmq this is preempted almost as a matter of style. For example, the
destructor pattern in CLASS, and also in zstr_free().  There's little
reason not to replace the calls to free() with a macro that also
nullifies the pointer after free()ing it.  Many project do this and for
good reason, as it eliminates a whole class of bugs.

The only downside I can see is that a double-free no longer triggers a
SEGV, because what is still a logical error is turned into a silent
no-op instead, and it may be the case that the SEGV would have alerted
us to other logical errors. But the caveat also applies to callers of
CLASS destructors and zstr_free(), so it's a trade-off which has already
been adopted to an extent.

### DIFF
--- a/src/czmq_classes.h
+++ b/src/czmq_classes.h
@@ -24,6 +24,8 @@
 //  External API
 #include "../include/czmq.h"
 
+#define FREE_AND_NULL(x) do {free(x); x = NULL;} while(0)
+
 //  Extra headers
 
 //  Opaque class structures to allow forward references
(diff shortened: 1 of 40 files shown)


bluca:
As the header in src/czmq_classes.h says, you've changed an autogenerated file, so this change will be reverted at the next regen

Please do the changes in https://github.com/zeromq/zproto and https://github.com/zeromq/zproject

jpitc:
https://github.com/zeromq/czmq/pull/1664

PR #1664, opened 2017-04-09
Problem: #1663 modified auto-generated file

jpitc:
Solution: Move changes to include/czmq.h and restrict FREE_AND_NULL
changes to non-generated code. Assuming generated code is correct, or
should be.

see https://github.com/zeromq/czmq/pull/1663#issuecomment-292722509

### DIFF
--- a/include/czmq_prelude.h
+++ b/include/czmq_prelude.h
@@ -419,6 +419,7 @@ typedef struct {
 
 #define streq(s1,s2)    (!strcmp ((s1), (s2)))
 #define strneq(s1,s2)   (strcmp ((s1), (s2)))
+#define freen(x) do {free(x); x = NULL;} while(0)
 
 //  Provide random number from 0..(num-1)
 //  Note that (at least in Solaris) while rand() returns an int limited by
--- a/src/czmq_classes.h
+++ b/src/czmq_classes.h
@@ -24,8 +24,6 @@
 //  External API
 #include "../include/czmq.h"
 
-#define FREE_AND_NULL(x) do {free(x); x = NULL;} while(0)
-
 //  Extra headers
 
 //  Opaque class structures to allow forward references
--- a/src/test_zgossip.c
+++ b/src/test_zgossip.c
@@ -13,8 +13,8 @@ assert_status (zactor_t *actor, int count)
             assert (atoi (status) == count);
             ready = true;
         }
-        FREE_AND_NULL (command);
-        FREE_AND_NULL (status);
+        freen (command);
+        freen (status);
     }
 }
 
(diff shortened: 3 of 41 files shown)


bluca:
Please don't, this is the public API header. Instead add the definition and usage in zproto and zproject

sappo:
The correct place for the define is `https://github.com/zeromq/czmq/blob/master/include/czmq_prelude.h`. Maybe as part of the `Inevitable macros` on line 418.

Regarding the name, I would prefer somethings a lot shorter without the caps e.g. `freen()` or `nfree()`.

sappo:
Also if this turn out to be a good thing, we should keep in mind to adjust the RFC.

bluca:
shorter name would be nice

but this is nothing specific to czmq, so if it goes in the zproject-generated header all projects can use it

sappo:
@bluca while I agree with your argument, currently CLASS project are by RFC definition depended on czmq and a lot of the defines in `czmq_prelude.h` are used by higher level projects, especially for platform independence. IMO putting the `FREE_AND_NULL` define in there would be the best short term solution. In long term we should try to port the `czmq_prelude.h` so that czmq won't be a hard dependency any more.

jpitc:
You both agree this would be a good thing, that's good. So:

1. Macro name
2. Which include file in the czmq tree should go into
3. Should the gsl in zproto and zproject be similarly modified

@bluca said:
> Please don't, this is the public API header. Instead add the definition and usage in zproto and zproject

ok, but you're mixing two separate issues. One has to do with czmq code itself, the other with the generated code in czmq (as you originally pointed out to me). 

@sappo said:
> Regarding the name, I would prefer somethings a lot shorter without the caps e.g. freen() or nfree().

ALLCAPS for macros is a very strong convention in c and I didn't/don't see any special reason to violate it here. Whatever name you like is fine by me.

regarding (3): In theory (hopefully practice) code generators create "perfect code" and the code is never manually modified. Arguably, that is a case where you **do** want logic errors (such as double frees) to explode instead of disarming them. My review of zproto/codec_c showed the changes are mostly limited to nullifying a pointer just before it goes out of scope. i.e., cases where inspection easily demonstrates correctness. 

If you can come to a consensus on these 3 points, I'll implement it.

sappo:
@jpitc thanks for your efforts so far:
1. The name is no concern for this PR.
2. When you add the define to czmq then place it in `czmq_prelude.h` 
3. The cleaner solution as @bluca mentioned is to add it to zproject to the projects `_library` file see (https://github.com/zeromq/zproject/blob/master/zproject_class.gsl#L42).
4. IMO zproto is out of scope for now.

jpitc:
- moved to `czmq_prelude.h`
- renamed macro to `freen()`, because lowercase macros seem to be the (unconventional) convention within czmq.

my 2c - as long as this pattern isn't officially what CLASS mandates it's not zproject's role to provide that macro publicly to every generated project and because zproto follows CLASS, the same reasoning applies.

If added to zproto before the rfc, my suggestion would be to declare `freen()` at the top of the generated .c file as a "private" macro and not in the public interface header (as @bluca stated), where people will inevitably start to depend on it directly in their own code.

PR #1670, opened 2017-05-02
Problem: zbeacon checks wrong UDP socket before send

bluca:
Solution: check self->udpsock_send which is actually used to send
rather than self->udpsock whcih is used just to receive


bluca:
Ping!

PR #1675, opened 2017-06-01
Problem: zhashx_purge() seems to leak when shrinking

jimklimov:
Solution: freen() the old self->items before reassigning to the newly allocated smaller one

### DIFF
--- a/src/zhashx.c
+++ b/src/zhashx.c
@@ -395,6 +395,7 @@ zhashx_purge (zhashx_t *self)
         size_t limit = primes [INITIAL_PRIME];
         item_t **items = (item_t **) zmalloc (sizeof (item_t *) * limit);
         assert (items);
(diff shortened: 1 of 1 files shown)


jimklimov:
Note: review desired, since this PR is more of an educated guess in debugging our component's issues (valgrind on Travis is ok against czmq-3.0.2 but reports multiple leaks against current master) - see e.g. https://travis-ci.org/42ity/fty-metric-composite/jobs/237155542 for details.

jimklimov:
Also posted a self-test PR separately at #1676 - hopefully it would expose the problem (and later confirm the fix) on Travis too :)

jimklimov:
yay, our project now passes with czmq4 and can deprecate requirement for czmq-3.0.2 :)

bluca:
Nice!

PR #1676, opened 2017-06-01
Problem: no selftest for zhashx_purge()

jimklimov:
Solution: make a test to see if it works well

### DIFF
--- a/src/zhashx.c
+++ b/src/zhashx.c
@@ -1303,6 +1303,23 @@ zhashx_test (bool verbose)
     assert (streq ((char *) zhashx_lookup (hash, "key2"), "Ring a ding ding"));
     zhashx_destroy (&hash);
 
+    //  Test purger and shrinker: no data should end up unreferenced in valgrind
+    hash = zhashx_new ();
+    assert (hash);
+    zhashx_set_destructor (hash, (zhashx_destructor_fn *) zstr_free);
+    zhashx_set_duplicator (hash, (zhashx_duplicator_fn *) strdup);
+    char valuep [255];
+    strcpy (valuep, "This is a string");
+    rc = zhashx_insert (hash, "key1", valuep);
+    assert (rc == 0);
+    strcpy (valuep, "Ring a ding ding");
+    rc = zhashx_insert (hash, "key2", valuep);
+    assert (rc == 0);
+    assert (streq ((char *) zhashx_lookup (hash, "key1"), "This is a string"));
+    assert (streq ((char *) zhashx_lookup (hash, "key2"), "Ring a ding ding"));
+    zhashx_purge (hash);
+    zhashx_destroy (&hash);
+
 #if defined (__WINDOWS__)
     zsys_shutdown();
 #endif
--- a/src/zhashx.c
+++ b/src/zhashx.c
@@ -1315,8 +1315,16 @@ zhashx_test (bool verbose)
     strcpy (valuep, "Ring a ding ding");
     rc = zhashx_insert (hash, "key2", valuep);
     assert (rc == 0);
+    strcpy (valuep, "Cartahena delenda est");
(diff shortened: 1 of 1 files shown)


jimklimov:
Note: until the merge of #1675 this test is expected to fail with valgrind; after that merge it is expected to pass.

jimklimov:
That's interesting... got an unrelated(?) hiccup in https://travis-ci.org/zeromq/czmq/jobs/238413291

````
596 * zsys: OK
597lt-czmq_selftest: src/zproc.c:1041: zproc_test: Assertion `stdout_read' failed.
598 * zproc: make[2]: *** [check-local] Aborted (core dumped)
599make[2]: Leaving directory `/home/travis/build/zeromq/czmq'
600make[1]: *** [check-am] Error 2
601make[1]: Leaving directory `/home/travis/build/zeromq/czmq'
602make: *** [check-recursive] Error 1
````

Its equivalent https://travis-ci.org/zeromq/czmq/jobs/238413291 from another PR passed some 10 minutes before. Is this a known intermittent error (and then is there an issue for it), or some new breakage induced by this selftest in a completely different area? :\

**Comment by bluca** [comment]
I've restarted the job to check, but I'm fairly sure the selftest here can't possibly affect zproc

**Comment by bluca** [comment]
@jimklimov it's not failing the valgrind run - I think it needs to have one more item to get a rehash:

if (self->size >= limit * LOAD_FACTOR / 100) 

With load_factor being 75 and limit being 3 (the first prime in the array), eg: 2.2 rounded down to 2 - on the first insert size is 0, on the second one is 1, and on a third one it will be 2 and cause a rehash, if I got things right

https://github.com/zeromq/czmq/blob/master/src/zhashx.c#L249

**Comment by jimklimov** [comment]
Bumped the code to add more list entries, though in https://github.com/zeromq/czmq/blob/master/src/zhashx.c#L395 the deciding factor seems to be 

    self->prime_index > INITIAL_PRIME

where `INITIAL_PRIME` is `0` per https://github.com/zeromq/czmq/blob/master/src/zhashx.c#L31 ...

This does not contradict your point though, as the list has to get extended, and I initially misread the clause, switching work with home chores :)

**Comment by jimklimov** [comment]
Wonder if we should go for cleanest setup - determine the limit (from arrays and math you posted) and populate that many plus some more entries, to survive any changes in tunables? Or is such hardcoding good enough?

**Comment by jimklimov** [comment]
Grrrr, passed again ;(

https://travis-ci.org/zeromq/czmq/builds/238475258

**Comment by bluca** [comment]
It does fail without the change:

```
==1311== 136 bytes in 1 blocks are definitely lost in loss record 1 of 1
==1311==    at 0x4C2DBC5: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1311==    by 0x4E613B3: safe_malloc (czmq_prelude.h:541)
==1311==    by 0x4E61918: s_zhashx_rehash (zhashx.c:210)
==1311==    by 0x4E61B24: zhashx_insert (zhashx.c:255)
==1311==    by 0x4E64AC2: zhashx_test (zhashx.c:1321)
==1311==    by 0x1094F7: test_runall (czmq_selftest.c:100)
==1311==    by 0x109A7F: main (czmq_selftest.c:209)
==1311== 
==1311== LEAK SUMMARY:
==1311==    definitely lost: 136 bytes in 1 blocks
==1311==    indirectly lost: 0 bytes in 0 blocks
==1311==      possibly lost: 0 bytes in 0 blocks
==1311==    still reachable: 0 bytes in 0 blocks
==1311==         suppressed: 0 bytes in 0 blocks
```

I think what's being run on Travis is the result of the merge, so it has the fix. If you back it out and ran make memcheck, you'll see this leak reported.

Thanks!

jimklimov:
Oh, good to know - thanks!

/me not a dummee ;)

PR #1679, opened 2017-06-04
Problem: parsing commandline is hard

vyskocilm:
Solution: add ligthweight and minimal API for accessing argv in the most
ergonomic, yet very flexible, way. Class zargs_t distinguish between
(positional) arguments and (named) parameters. And provide different
czmq compatible API for accessing those.

It aims to be getopt compatible, so comes with -- as a way to
distinguish between arguments and parameters. In future it will handle
the --with=value and -Wvalue variants as well.

Note: this is NOT a replacement for declarative way of defining command line api (like getopt, argparse, or doctopt). This is simpler, platform independent, CLASS/czmq comaptible API for accessing your argv.

# Pull Request Notice

```
Problem: X is broken

Solution: do Y and Z to fix X
```

Please avoid sending a pull request with recursive merge nodes, as they
are impossible to fix once merged. Please rebase your branch on
zeromq/czmq master instead of merging it.

git remote add upstream git@github.com:zeromq/czmq.git
git fetch upstream
git rebase upstream/master
git push -f

In case you already merged instead of rebasing you can drop the merge commit.

git rebase -i HEAD~10

Now, find your merge commit and mark it as drop and save. Finally rebase!

If you are a new contributor please have a look at our contributing guidelines:
[CONTRIBUTING.md](https://github.com/zeromq/czmq/blob/master/CONTRIBUTING.md)


vyskocilm:
cc @sappo  @bluca please review the API, however I hope it makes a sense!

project.xml
```diff
@@ -13,10 +13,10 @@
     <use project = "libzmq" />
     <use project = "uuid" optional = "1" implied = "1" />
     <use project = "systemd" optional = "1" implied = "1" min_major = "200" />
-    <target name = "*" />
```
bluca: any particular reason to remove this?

builds/gyp/project.gyp
```diff
@@ -138,7 +140,6 @@
         '../../src/czmq_classes.h'
       ],
       'dependencies': [
-        '../../../libzmq/builds/gyp/project.gyp:libzmq',
```
bluca: This was probably done by zproject, don't think it's right to remove it?

bluca:
Looks great, just noticed two minor things that were probably caused by zproject

project.xml
vyskocilm: Yes, my zproject rpm was broken ... I did drop it in order to continue with a development. Will bring it back.

vyskocilm:
@bluca thanks for spotting it - I put it back and did gsl project.xml. As I am using zproject from OBS, I believe this is way to go on Linux

bluca:
Great, thanks! The API looks fine to me.

There seems to be a problem with Python bindings though: https://travis-ci.org/zeromq/czmq/jobs/239377425

bluca:
Seems strange, not sure why Python is complaining about zargs_lookup, doesn't seem to be anything out of the ordinary

bluca:
Ah I see the problem, 2 APIs have the name set to zargs_lookup[x] in the API: https://github.com/zeromq/czmq/blob/master/api/zargs.api#L79
So that's how they get generated and imported and exported, but the actual implementation calls them zargs_param_lookup[x]: https://github.com/zeromq/czmq/blob/master/src/zargs.c#L187

So the zargs_lookup[x] symbols do not exist in the shared library since there's no function with those names, and thus Python fails to import them

Could you please fix it?

vyskocilm:
https://github.com/zeromq/czmq/pull/1680

Issue #1684, opened 2017-06-13
zsys access from python


diorcety:
There is no way to access zsys_* functions from python binding

bluca:
Yes that is a known problem, since they are not generated by zproject there's no binding generation. No idea how we can solve it.

diorcety:
This trigger an issue on Windows. Either a crash or a stall (after last python code)

zsys_shutdown is called atexit.
But something is done before atexit by python internal code (WSACleanup?)
This lead to this issue: https://lists.zeromq.org/pipermail/zeromq-dev/2016-March/030059.html

if i call 
    import ctypes
    lib = ctypes.cdll.LoadLibrary('libczmq')
    lib.zsys_shutdown.restype = None
    lib.zsys_shutdown.argtypes = []
    lib.zsys_shutdown()
at the end of my python program, i have no issue at all...

bluca:
I thought there was no atexit on windows?

bluca:
That was with a DLL build though, might not be your case:
https://github.com/zeromq/czmq/pull/1608
https://github.com/zeromq/zproject/pull/836

jemc:
> Yes that is a known problem, since they are not generated by zproject there's no binding generation. No idea how we can solve it.

Is there a deficiency in zproject or oddity about `zsys` that prevents someone writing the zproject API def for it? Since there is no `zsys` class instance, I think that just means all of the methods would be `singleton` methods, as described here:

https://github.com/zeromq/zproject/blob/dfb1af34e52350cd6a94956f618e41611c8eb152/api/myclass.api#L99

bluca:
Honestly not sure. Perhaps it's simply that nobody has tried to convert it until now. PRs welcome as usual :-)

jemc:
Right, I just wanted to see if there was a known deficiency in the zproject API model, since I helped design it (however many years ago).

jemc:
So @diorcety if you wanted to try your hand at writing an API def for `zsys` to go in [this directory](https://github.com/zeromq/czmq/tree/master/api), you could generate the bindings you crave - and let us know if you get stuck and need to ask questions about how it works.

diorcety:
I manage do create the api. but i have two issues. First one:
`(zproject_nodejs.gsl 704) Undefined expression:`
Second one: Even if I skip nodejs, python binding is not built.

sappo:
Push your changes to your fork and I'll have a look at it.

[...]

PR #1685, opened 2017-06-15
Introduce zsys.api

diorcety:
Use api file in order to support zsys in different bindings. cf #1684
The only api change is zsys_file_size return type.
 zsys_file_size return type is ssize_t
stat_buf.st_size type is off_t
so i replaced zsys_file_size return type by off_t which is file_size type in the api file


bluca:
Sorry, but we cannot make backward incompatible changes to the public API.

Is there any way to keep zsys_file_size the same?

bluca:
Keeping its definition outside of the API file (leave it in the header along with the zsys_interrupted definitions) would be fine IMHO, given it's never been available in the bindings anyway

jemc:
@diorcety Great work!

To avoid changing the signature of `zsys_file_size` as @bluca has asked, zproject needs to be made aware of `ssize_t` as a new supported type.

bluca:
Returning off_t is actually the right thing as it's forward compatible with Large File Support:

https://lintian.debian.org/tags/binary-file-built-without-LFS-support.html

IMHO it's better to just keep that one off the API rather than adding deprecated stuff to zproject

PR #1694, opened 2017-06-26
fixed Werror messages building on mingw32 mingw64

opunix:
Hi,

I've fixed the wanings errors which i got during the compilation of czmq on mingw64 ... it seems like the check for mingw64 did not worked very well ... because of the host_os switch condition .. it searches for mingw64 in the host_os variable but it is in my case x86_64-w64-mingw32. That has the side effect that it runs the mingw32 settings ... I'm not sure if it should be changed or added a new case so i decided to just set the same FLAGS to mingw32 and mingw64 as a more generic solution.

BR Frank :-)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -477,6 +477,8 @@ case "${host_os}" in
         AC_DEFINE(CZMQ_HAVE_HPUX, 1, [Have HPUX OS])
         ;;
     *mingw32*)
+	CFLAGS="-Wno-error=format -Wno-unused-function -Wno-unused-variable -D_XOPEN_SOURCE $CFLAGS"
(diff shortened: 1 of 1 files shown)


bluca:
Thanks, but as the top of that file says it's autogenerated and those changes will be overwritten.

Could you please send the changes to this file and repo instead? https://github.com/zeromq/zproject/blob/master/zproject_autotools.gsl

Also please use only spaces for indentation

opunix:
OK,

I created a pull request here:

https://github.com/zeromq/zproject/pull/928

BR Frank :-)

bluca:
Thanks, closing as it's included in the next regen: https://github.com/zeromq/czmq/pull/1695

sappo:
Neat 👍

PR #1702, opened 2017-07-04
changed description of 2 zhashx functions

ju-emb:
two corrections on descriptions in zhashx

one change is to correct an error in description for set_key_hasher.
bluca told me on IRC how it has to be right.

other is an addition to the description of set_key_comparator function
to eliminate possible confusion what the callback function has to return.

### DIFF
--- a/api/zhashx.api
+++ b/api/zhashx.api
@@ -278,8 +278,8 @@
     </method>
 
     <method name = "set_key_hasher">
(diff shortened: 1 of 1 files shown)


bluca:
Thanks!

For your next PRs, please use the "Problem: x Solution: y" format for the commit messages, as the template suggests

PR #1707, opened 2017-07-19
Problem: can't use zproto non-virtual API for zactor pipe

vyskocilm:
Solution: add a custom destructor function callback, which overrrides default
zmsg_t with $TERM by own message format. With this patch it's possible to use
custom messages built by zproto_codec_c without converting them from/to zmsg_t.

Note the actor itself MUST handle new termination message, otherwise it won't be stopped

### DIFF
--- a/src/zactor.c
+++ b/src/zactor.c
@@ -42,6 +42,7 @@
 struct _zactor_t {
     uint32_t tag;               //  Object tag for runtime detection
     zsock_t *pipe;              //  Front-end pipe through to actor
+    zactor_destructor_fn *destructor;   //  Custom destructor for actor, sends $TERM by default
 };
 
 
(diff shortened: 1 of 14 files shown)


src/zactor.c
```diff
@@ -267,6 +284,24 @@ echo_actor (zsock_t *pipe, void *args)
...
 }
 
+static void
+KTHXBAI_actor (zsock_t *pipe, void *args) {
+
+    zsock_signal (pipe, 0);
+    while (!zsys_interrupted) {
+        char *str = zstr_recv (pipe);
+        if (streq (str, "$KTHXBAI"))
+            break;
```
bluca: this leaks str

bluca:
Nice stuff. But should the new APIs be added as DRAFT for a while?

vyskocilm:
Yes, no problem

Dne 19. 7. 2017 5:59 odpoledne napsal uživatel "Luca Boccassi" <
notifications@github.com>:

bluca:
Ok, I'll send a followup

bluca:
https://github.com/zeromq/czmq/pull/1708

PR #1710, opened 2017-07-21
problem: zframe_meta documentation is incorrect

JustinAzoff:
zframe_meta returns the value from zmq_msg_gets directly and according
to http://api.zeromq.org/4-2:zmq-msg-gets

The zmq_msg_gets() function shall return the string value for the
property if successful. Otherwise it shall return NULL and set errno to
one of the values defined below. The caller shall not modify or free the
returned value, which shall be owned by the message. The encoding of the
property and value shall be UTF8.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2463,7 +2463,8 @@ This is the class interface:
         zframe_data (zframe_t *self);
     
     //  Return meta data property for frame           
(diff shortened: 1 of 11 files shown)


JustinAzoff:
No updates to the tests, the tests are already correct and do not free the returned string.

PR #1712, opened 2017-07-22
Problem: zsock_brecv doc is wrong w.r.t buffer management

bluca:
Solution: clarify what to do for each supported type and regenerate


vyskocilm:
That's the best answer you can make 😀

PR #1713, opened 2017-07-22
problem: no way to loop over all certs in a zcertstore

JustinAzoff:
solution: add zcertstore_certs method that returns a zlistx of certs

This may not be 100% right yet - I'm not sure how to regenerate the bindings and documentation.  The goal is to be able to do something like this:

    zcert_t *cert;
    zlistx_t *certs = zcertstore_certs(certstore);
    cert = (zcert_t *) zlistx_first(certs);
    while (cert) {
        char *endpoint = zcert_meta (cert, "endpoint");
        if(endpoint) {
            fprintf(stderr, "Connect to %s using %s\n", endpoint, zcert_public_txt(cert));
        }
        cert = (zcert_t *) zlistx_next(certs);
    }
    zlistx_destroy(&certs);

That way an application won't require any configuration aside from a certstore path.  This could also be used to find the cert in a certstore that has the private key as well.

### DIFF
--- a/api/zcertstore.api
+++ b/api/zcertstore.api
@@ -65,6 +65,11 @@
         Print list of certificates in store to logging facility
     </method>
 
+    <method name = "certs" state = "draft">
+        Return a list of all the certificates in the store
+        <return type = "zlistx" fresh = "1" />
+    </method>
+
     <method name = "test" singleton = "1">
         Self test of this class
         <argument name = "verbose" type = "boolean" />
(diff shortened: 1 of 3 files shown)


bluca:
Here's the regen:
https://github.com/zeromq/czmq/pull/1714

If you are running linux you can install zproject from our OBS repos: https://build.opensuse.org/project/show/network:messaging:zeromq:git-stable and then you just run ```gsl project.xml``` in the czmq directory

JustinAzoff:
Got it, thanks!

PR #1716, opened 2017-07-22
problem: zcertstore_certs was broken

JustinAzoff:
solution:

zcertstore configures itself with a destructor that calls
zcert_destroy.  zhashx_values copies this destructor by default, causing
zlistx_destroy on the returned list to incorrectly destroy the
certificates.  To fix this the destructor must be disabled on the
zlist_x before it can be destroyed.

### DIFF
--- a/api/zcertstore.api
+++ b/api/zcertstore.api
@@ -66,7 +66,10 @@
     </method>
 
     <method name = "certs" state = "draft">
-        Return a list of all the certificates in the store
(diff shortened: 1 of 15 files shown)


JustinAzoff:
Sorry I didn't catch this on the first go around, the test had the wrong #if and I wasn't calling zlistx_destroy in the test anyway.  I did call zlistx_destroy in my application which promptly crashed when a new cert was added :-)  It's working much better now.

PR #1718, opened 2017-08-01
Problem: can not see why zsys_test fails

jimklimov:
Solution: augment the timing-dependent portion of code and test with optional verbosity (requested privately from tests, invisible for ordinary consumers)

Context:

As discussed on IRC, I have this test failing in valgrind of czmq-master built as part of a larger project on Travis. I could not reproduce the issue isolated from that build, nor in local rebuilds, so its heaviness and/or Travis virtualization may be a contributing factor (this test relies on speed that a recently created file is "young", so lags are fatal). Curiously, non-memcheck builds against czmq-master, as well as any sort of builds against our legacy baseline fork of czmq-v3.0.2 (where these routines seem to be same as in master) do not fail the same tests in Travis.

In order to debug this better, I added some optional tracing to this part of the code, as agreed with @bluca .

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -673,8 +673,9 @@ zsys_file_delete (const char *filename)
 //  --------------------------------------------------------------------------
 //  Check if file is 'stable'
 
+// Internal implementation rigged with debugs and called from selftest
 bool
-zsys_file_stable (const char *filename)
+s_zsys_file_stable (const char *filename, bool verbose)
 {
     struct stat stat_buf;
     if (stat (filename, &stat_buf) == 0) {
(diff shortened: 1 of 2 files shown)


src/zsys.c
```diff
@@ -673,24 +673,43 @@ zsys_file_delete (const char *filename)
 //  --------------------------------------------------------------------------
 //  Check if file is 'stable'
 
+// Internal implementation rigged with debugs and called from selftest
 bool
-zsys_file_stable (const char *filename)
+s_zsys_file_stable (const char *filename, bool verbose)
```
bluca: Please add "static" otherwise it will get exported

jimklimov: Thanks, posted #1720

PR #1719, opened 2017-08-01
Problem: CZMQ master build fails in JNI binding

jimklimov:
Solution: do not use a reserved Java keyword "interface" in Zsys.java; rename function to getInterface()

Note: maybe there should be some place to document this modification and different behavior compared to other bindings. This change likely breaks no consumers because previous code never passed the build since introduction, per Travis build history (see `git diff f05c010 1042d8f` for change between green and red states). Probably the binding code is generated from the C code - if so, the parser script should also be amended to produce this exclusion and perhaps comment upon re-generations.

### DIFF
--- a/bindings/jni/src/main/java/org/zeromq/czmq/Zsys.java
+++ b/bindings/jni/src/main/java/org/zeromq/czmq/Zsys.java
@@ -380,9 +380,11 @@ public class Zsys {
     }
     /*
     Return network interface to use for broadcasts, or "" if none was set.
+    Note that "interface" is among reserved Java keywords, so it had to be
+    renamed here manually to "getInterface".
     */
     native static String __interface ();
-    public String interface () {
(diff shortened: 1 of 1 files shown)


sappo:
This will be overridden with the next regen :(

sappo:
I opened an issue in zeromq/zproject#947

jimklimov:
Yep, that's why I wrote my comment in the PR description for probable follow-up tasks.

Also, at least I myself usually `git diff` the results of regeneration, in no small part to identify such manual(-looking) differences in unexpected places, to cherry-pick them into the next life of the project or to amend the generator or to refactor the project. Either way, to take a conscious choice and action.

For example, following C4, a "red" Travised commit would not get merged - hence pure blind regen would not break it. Would have to be fixed back to pass :)

PR #1724, opened 2017-08-01
Problem: zsys_test can fail if old run artifacts are present

jimklimov:
Solution: Drop dir in zsys_test(). It seems safe to just do so, no extra checks on presence are needed.

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -1948,6 +1948,11 @@ zsys_test (bool verbose)
     zsys_file_mode_private ();
     if (verbose)
         printf ("zsys_test() at timestamp %" PRIi64 ": "
+            "Dropping .testsys/subdir (may be absent)\n",
+            zclock_time() );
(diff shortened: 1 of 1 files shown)


jimklimov:
Ok, this was a bit naïve - the directory is only deleted if empty. So we have to drop both, or make some recursive-delete routine (did I miss one already present?)

Posted follow-up PRs to attack this from two angles - clean up the selftest code in zfile and zsys sources, to try and delete FS objects that are expected to be absent at start of test and whose presence can corrupt test results, and revise the `SELFTEST_DIR_RW` usage in autotools at least (which is what we use and know how to code) so that the RW directory is wiped before tests, and is checked for emptiness if a selftest run has succeeded (so a test which did not clean up after itself is a failed test). The latter solution does not cover the usecases where autotools/make are not used to execute the test (e.g. developers run code in their workspaces, or use cmake, or ...) hence the former cleaner but harder solution is proposed too :)

PR #1725, opened 2017-08-01
problem: missing zcert_new_from_txt function 

wesyoung:
solution: create zcert_new_from_txt function that takes in keys as text and generates a cert (instead of just from binary or disk)

added as a DRAFT function for now... thought about modifying new_from but wasn't sure if that would potentially break things for others...

### DIFF
--- a/src/zcert.c
+++ b/src/zcert.c
@@ -98,6 +98,34 @@ zcert_new_from (const byte *public_key, const byte *secret_key)
     return self;
 }
 
+#ifdef CZMQ_BUILD_DRAFT_API
+//  --------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


src/zcert.c
```diff
@@ -98,6 +98,34 @@ zcert_new_from (const byte *public_key, const byte *secret_key)
     return self;
 }
 
+#ifdef CZMQ_BUILD_DRAFT_API
```
bluca: you don't need to wrap the implementation in the ifdef - it's exported as private by default

wesyoung: aye, but prob still for the zcert_test block, right?

bluca: yes

PR #1729, opened 2017-08-02
Problem: zsys_file_stable() check for (age>1000) is weak

jimklimov:
Solution: Document the analysis of why checking for (age>1000) when we account in values rounded to thousands is misleading, propose better solutions and an improvement for current implementation.

This takes off from discussion in IRC.

### DIFF
--- a/api/zsys.api
+++ b/api/zsys.api
@@ -303,6 +303,22 @@
         <return type = "integer" />
     </method>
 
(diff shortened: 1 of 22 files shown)


bluca:
All tests fail in zdir, could you please have a look?

jimklimov:
Yes, on it. It is about those obscure polls to wait for 1001 msec we discussed yesterday, I guess.

PR #1731, opened 2017-08-03
Problem: once in a few million runs, randof() still lies

jimklimov:
Solutions: in czmq_prelude.h randof() :
* make sure to use RAND_MAX only where available;
* make more effort to multiply NUM by a factor of less than 1.0 - limit the chosen random function result by chosen maximum minus one (remainder of...) in all cases, and parenthesize calculation of the "0...<1" factor to avoid overflows etc.
* do not calculate the random-factor with large numbers like UINT32_MAX or RAND_MAX (big on e.g. Linux-64bit), because either `float` can not discern these values plus-minus several discrete levels at all, or (in experiment with `double`) the resulting factor has too many nines and multiplication yields "num.0000000" anyway. Constrain this to smaller numbers, use UINT16_MAX which passed the artificial tests ok.

Details:
During builds of previous PR and master branches, which added a loop of 10 million executions of `randof()` asserting the results, about 5-6 of 18 jobs executed failed right in this test with:
````
lt-czmq_selftest: ../src/zhashx.c:1300: zhashx_test: Assertion `testnbr != testmax' failed.
```

This means that due to some hardware (including VM) floating point magic, casts, libc/libmath or whatever other reasons, the macro sometimes ends up multiplying `num` by `1.0` or some factor close enough that casting back to `int` produces the `num` argument value rather than expected maximum of `num-1`.

Given that generally the `num` token can be some `function()` call, so we can not refer to it twice in the macro to doctor the outputs, e.g. adding a `% num` to limit the value. Using `num - 1` can yield unexpected negative outputs. So short of converting to function, there is not much more refinement in macro that I can think of at the moment...

### DIFF
--- a/include/czmq_prelude.h
+++ b/include/czmq_prelude.h
@@ -421,21 +421,22 @@ typedef struct {
 #define strneq(s1,s2)   (strcmp ((s1), (s2)))
 #define freen(x) do {free(x); x = NULL;} while(0)
 
-//  Provide random number from 0..(num-1)
+//  randof(num) : Provide random number from 0..(num-1)
 //  Note that (at least in Solaris) while rand() returns an int limited by
 //  RAND_MAX, random() returns a 32-bit value all filled with random bits.
 //  The math libraries on different platforms and capabilities in HW are a
 //  nightmare. Seems we have to drown the code in casts to have reasonable
-//  results...
+//  results... Also note that the 32-bit float has a hard time representing
+//  values close to UINT32_MAX that we had before, so now limit to UINT16_MAX.
 #if defined(RAND_MAX)
 # if (defined (__WINDOWS__)) || (defined (__UTYPE_IBMAIX)) \
  || (defined (__UTYPE_HPUX)) || (defined (__UTYPE_SUNOS)) || (defined (__UTYPE_SOLARIS))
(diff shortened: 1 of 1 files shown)


jimklimov:
Code for staging the experiment to strain the limits:

````
#include <stdlib.h>
#include <stdio.h>
#include <math.h>

#ifndef MAXVAL
#define MAXVAL UINT16_MAX
//#define MAXVAL UINT32_MAX
//#define MAXVAL INT32_MAX
//#define MAXVAL RAND_MAX
//#define MAXVAL 2147483647
#endif

main() {
    float top = (float)MAXVAL - 1.0;
    float bot = (float)MAXVAL + 1.0;
    int num = 999;
    float factor = top / bot;
    float fnum = (float)(num) * factor;
    float ffnum = floorf (fnum);
    int inum = (int)(ffnum);
    printf ("   num= %d\n   top= %.17f\n   bot= %.17f\nfactor= %.17f\n  fnum= %.17f\n ffnum= %.17f\n  inum= %d\nMAXVAL= %jd\n",
        num, top, bot, factor, fnum, ffnum, inum, (intmax_t)MAXVAL);
}
````

and run with e.g. `gcc -o xxxx xxxx.c -lm -DMAXVAL=INT16_MAX && ./xxxx` to simply test different `MAXVAL`s.

PR #1732, opened 2017-08-03
Problem: File stable timeout tests still fail

jimklimov:
Solutions:
* use longer default timeout that should suffice for laggy hosts and OSes (note: so perhaps also more incentive to tune it via `configure` for host OSes with better known characteristics and/or extended stat-struct fields with better timestamp resolution); 
* use even longer poll-timeouts in zdir_test() waiting for FS changes to be detected


jimklimov:
Travis fails so far are due to randof(), tackled in another PR.

One Windows run failed in zsock test, no idea what that was...

PR #1735, opened 2017-08-03
Problem: A few more components do not use SELFTEST_DIR

jimklimov:
Solution: Enhance zdir.c and zdir_patch.c tests with SELFTEST_DIR_RW support


jimklimov:
Note: travis lags today, but my branch almost finished building - all green so far: https://travis-ci.org/jimklimov/czmq/builds/260658137

PR #1736, opened 2017-08-03
Problem: zproc_test() can not find its zsp helper

jimklimov:
Solution: figure out and unravel the draft :)

Note: this only pops up in builds with `--enable-drafts=yes` as `zproc` is still experimental.

Merge criteria: the issue is fixed if such errors are not seen in a run:
````
 * zproc: W: (czmq_selftest) 17-08-03 17:13:56 cannot detect zsp binary, src/zsp does not exist
SKIPPED (zsp helper not found)
````

but rather success is logged:
````
 * zproc: OK
````

PR #1737, opened 2017-08-03
Problem: zproc_test() still can not find its zsp helper

jimklimov:
Solution:
* figure out and unravel the draft :)
* fix a side effect of zsys tests (changed CWD)
* make this fix compile on Windows too

Continuation of #1736

Note: this only pops up in builds with `--enable-drafts=yes` as `zproc` is still experimental, and only against ZMQ4+.

Merge criteria: the issue is fixed if such errors are not seen in a run:
````
 * zproc: W: (czmq_selftest) 17-08-03 17:13:56 cannot detect zsp binary, src/zsp does not exist
SKIPPED (zsp helper not found)
````

but rather success is logged:
````
 * zproc: OK
````

### DIFF
--- a/src/zdir.c
+++ b/src/zdir.c
@@ -1025,7 +1025,7 @@ zdir_test (bool verbose)
 
     // wait for initial file to become 'stable'
 #ifdef CZMQ_BUILD_DRAFT_API
-    zclock_sleep (zsys_file_stable_age_msec() + 50);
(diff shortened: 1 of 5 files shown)


jimklimov:
Seems ok to merge : windows built, travis built, report matches expectations:
````
897 * zproc: zproc_test() : current working directory is /home/travis/build/jimklimov/czmq
898 OK
````

Later commit fixed this to not spam with reported CWD if all goes well.

PR #1738, opened 2017-08-03
Problems: Travis using 12.04, and OSX builds are slow

bluca:
Switch to Trusty and avoid installing binutils and ossp-uuid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,7 +97,7 @@ addons:
     - libzmq3-dev
 
 before_install:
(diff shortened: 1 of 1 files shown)


jimklimov:
Is this a typo? Trusty, 16.04 and 12.04?.. Or should there appear a new 16.04 repo for ZMQ products?

bluca:
We do have repos but 16.04, but yes it's a typo - Travis runs on 14.04

PR #1741, opened 2017-08-05
Problem: randof() values are too discrete

jimklimov:
Solutions:
* make it verifiable by experiment
* make `randof()` implementation modular to simplify experiments and adaptation to platforms with special needs
* make `randof()` performance fuzzier so we get more different random numbers (within specified range)

After recently limiting the range of values in discrete floating-point math for `randof()` in #1730 et al, I continued testing and found that the spectrum of returned values is noticeably discrete as we raise the range of values we want to pick. This makes sense, as the amount of random values is limited by the chosen MAX which we use to limit the bit-size of numbers and so take remainder-of division.

This PR adds the tool I made to test different variants, and a tweakable reimplementation of the `randof()` macro as a number of reusable components. In particular, to see the impact of math with different precision on different hardware it is now possible to use alternate `ZSYS_RANDOF_FLT` (defaults to `float`) during compilation, or try out some other `ZSYS_RANDOF_FUNC` if available on a particular platform. It is expected that future contributions of special care for special platforms, if needed, will become more simple in this code.

This PR also limits the `ZSYS_RANDOF_MAX` with a greater default value found per experiment (`2^26 - 1` caused no collisions through `int`~`float` conversions over billions of loops, while slightly larger values did).

Finally, the `randof()` macro fuzzies the returned spectrum a bit, leading to colossal improvement in quality of random results even where the range of numbers is larger than the safe (or chosen) `ZSYS_RANDOF_MAX` value, at the expense of smaller probability to reach as close to `1.0` factors as was possible previously (while still staying strictly under `1.0`) and a small hit to performance (twice the `random()` calls).

With fuzziness and relatively few iterations:
````
I: 17-08-05 04:00:09 test_randof : Running random loop for 3000000000 iterations and \
    99999999 values (factor of 30)
I: 17-08-05 04:08:02 test_randof : Testing hit-counts of 99999999 entries
I: 17-08-05 04:08:02 test_randof : Got 33847 zero-entries (misses) from 99999999 range \
    (0.03385 %) with ZSYS_RANDOF_MAX=49999999 and float precision and random()
````

Without fuzziness:
````
I: 17-08-05 03:11:40 test_randof : Running random loop for 3000000000 iterations and 99999999 \
    values (factor of 30)
I: 17-08-05 03:16:20 test_randof : Testing hit-counts of 99999999 entries
I: 17-08-05 03:16:20 test_randof : Got 73536433 zero-entries (misses) from 99999999 range \
    (73.54 %) with ZSYS_RANDOF_MAX=49999999 and float precision and random()

test-randof: ./src/test_randof.c:169: main: Assertion `(rndmisses < (testmax / 3 ))' failed.
````

````
I: 17-08-05 03:17:07 test_randof : Running random loop for 3000000000 iterations and 48999999 \
    values (factor of 61.2245)
I: 17-08-05 03:21:16 test_randof : Testing hit-counts of 48999999 entries
I: 17-08-05 03:21:16 test_randof : Got 25431729 zero-entries (misses) from 48999999 range \
    (51.9 %) with ZSYS_RANDOF_MAX=49999999 and float precision and random()
````

There are even more stacked `random()` calls when the `ZSYS_RANDOF_MAX` is very small compared to expected `num`, giving bearable results even with large ranges thousands of times wider than this MAX:
````
I: 17-08-05 04:34:51 test_randof : Running random loop for 300000000 iterations and 9999999 \
    values (factor of 30)
I: 17-08-05 04:35:55 test_randof : Testing hit-counts of 9999999 entries
I: 17-08-05 04:35:55 test_randof : Got 2140945 zero-entries (misses) from 9999999 range \
    (21.41 %) with ZSYS_RANDOF_MAX=29999 and float precision and random()
````

Same test with numbers as long as above - note the impact on timing, otherwise same miss percentage:
````
I: 17-08-05 04:39:45 test_randof : Running random loop for 3000000000 iterations and 99999999 \
   values (factor of 30)
I: 17-08-05 04:52:40 test_randof : Testing hit-counts of 99999999 entries
I: 17-08-05 04:52:40 test_randof : Got 21587571 zero-entries (misses) from 99999999 range \
    (21.59 %) with ZSYS_RANDOF_MAX=29999 and float precision and random()
````

There are practical limits to this approach. so `ZSYS_RANDOF_MAX` on the order of 1000 missed over 90% of the range with limits above. Don't use such big numbers on such small systems :)

And ultimately, I added a test to use this tool in `ci_build.sh` after all the `distcheck`s are complete, to give even more runs to this code under different conditions and on different platforms (e.g. to make sure we do not ever get `1.0` or greater factors, and that dispersion stays reasonable).


bluca:
This is some great stuff, thanks!

If you want to get data from Windows the test run has to be added to the cmake ci-build.sh

PR #1742, opened 2017-08-06
Problem: test-randof not used in CMake

jimklimov:
Solution: add test-randof to cmake/ci_build.sh as suggested in #1741 (to cover more platforms) and apply some other improvements from main script


bluca:
perhaps a plugin-system like what we have for cmake/automake?

bluca:
test -f ./ci_build_plugin.sh && source ./ci_build_plugin.sh

jimklimov:
Do you mean for the custom changes? The problem is that such changes can be man-made in random places of code, not necessarily where the plugin-sourcer would be. But still, an option for at least some cases...

bluca:
We could start with a post-build hook, and if it becomes a necessity add a pre-build one

PR #1743, opened 2017-08-06
Problem: test-randof not called in Windows CI

jimklimov:
Solution: add the call to appveyor tests

### DIFF
--- a/src/test_randof.c
+++ b/src/test_randof.c
@@ -141,8 +141,11 @@ int main (int argc, char *argv [])
     assert (rndcnt);
     memset (rndcnt, 0, sizeof (size_t) * testmax);
 
-    zsys_info ("test_randof : Running random loop for %jd iterations and %jd values (factor of %.6Lg)",
-        itermax, testmax, (long double)itermax/(long double)testmax);
+    unsigned int random_seed = (unsigned int) zclock_time ();
+    srandom (random_seed);
+
+    zsys_info ("test_randof : Running random loop for %jd iterations and %jd values (factor of %.6Lg) and random seed %u",
+        itermax, testmax, (long double)itermax/(long double)testmax, random_seed);
     for (iteration = 0; iteration < itermax && !zsys_interrupted ; iteration++) {
         testnbr = randof (testmax);
         assert (testnbr != testmax);
--- a/src/test_randof.c
+++ b/src/test_randof.c
@@ -126,12 +126,18 @@ int main (int argc, char *argv [])
         return 1;
     }
 
+    // WARNING: Code below implies that "size_t" used by zmalloc() and memset()
+    // would suffice to count the testmax-based byte size on your platform
+#ifdef SIZE_MAX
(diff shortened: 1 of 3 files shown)


src/test_randof.c
```diff
@@ -126,17 +126,26 @@ int main (int argc, char *argv [])
...
 
-    size_t *rndcnt = (size_t*)zmalloc (sizeof (size_t) * testmax);;
+    size_t *rndcnt = (size_t*)zmalloc (sizeof (size_t) * testmax);
     assert (rndcnt);
     memset (rndcnt, 0, sizeof (size_t) * testmax);
 
-    zsys_info ("test_randof : Running random loop for %jd iterations and %jd values (factor of %.6Lg)",
-        itermax, testmax, (long double)itermax/(long double)testmax);
+    unsigned int random_seed = (unsigned int) zclock_time ();
+    srandom (random_seed);
```
bluca: shouldn't this be srand if it's being used with srand? one is posix the other is stdlib

src/test_randof.c
```diff
@@ -126,17 +126,26 @@ int main (int argc, char *argv [])
...
         zsys_info ("test_randof - development helper for randof() testing");
 
-    size_t *rndcnt = (size_t*)zmalloc (sizeof (size_t) * testmax);;
+    size_t *rndcnt = (size_t*)zmalloc (sizeof (size_t) * testmax);
     assert (rndcnt);
     memset (rndcnt, 0, sizeof (size_t) * testmax);
 
-    zsys_info ("test_randof : Running random loop for %jd iterations and %jd values (factor of %.6Lg)",
-        itermax, testmax, (long double)itermax/(long double)testmax);
+    unsigned int random_seed = (unsigned int) zclock_time ();
```
bluca: in libzmq we also add the pid:

```
#if defined ZMQ_HAVE_WINDOWS
    int pid = (int) GetCurrentProcessId ();
#else
    int pid = (int) getpid ();
#endif
srand ((unsigned int) (clock_t::now_us () + pid));
```

src/test_randof.c
jimklimov: I looked around czmq code - there is a `srandom` macro that is mapped to `srand` on Windows, and a number of places initialize with `srandom(time(NULL))` or similar (different) spells.

src/test_randof.c
jimklimov: I guess there is a case for `seed_randof()` macro ;)

src/test_randof.c
bluca: ah perfect, missed that

jimklimov: That said, I think the code makes quite a bit of assumptions based on macros rather than `configure` (or equivalent) results on the actual build system. Well, it did not explode so far :)

bluca: Yes indeed, the whole prelude thing should move to zproject and be handled by the build system, it's on my very long todo :-)
There's also the problem that it's in a public header so it will have to stay there as well

PR #1753, opened 2017-09-04
Update zloop.c

Crossverse:
Code mistake


c-rack:
Thanks! ❤️ 

For your next pull request, could you please use our "Problem / Solution" pattern?

> Before sending a pull request make sure each commit solves one clear, minimal, plausible problem. Further each commit should have the following format:

```
Problem: X is broken

Solution: do Y and Z to fix X
```

See https://rfc.zeromq.org/spec:42/C4/ in section 2.3 for details.

PR #1757, opened 2017-09-11
problem: gossip missing curve support

wesyoung:
solution: add initial curve support to zgossip


bluca:
It looks like there are a few problems (fails to build on Linux, tests segfault on Windows), do you want to fix them before merge?

wesyoung:
ya. will take a look.

bluca:
Ok, thanks - then I'll post a couple of suggestions as well

src/zgossip.c
```diff
@@ -114,6 +114,11 @@ struct _server_t {
 
     tuple_t *cur_tuple;         //  Holds current tuple to publish
     zgossip_msg_t *message;     //  Message to broadcast
+
+#ifdef CZMQ_BUILD_DRAFT_API
```
bluca: this structure, and the static function as well, are internal only so for simplicity there's no need to wrap them in ifdefs

src/zgossip.c
```diff
@@ -274,7 +292,12 @@ server_method (server_t *self, const char *method, zmsg_t *msg)
     if (streq (method, "CONNECT")) {
         char *endpoint = zmsg_popstr (msg);
         assert (endpoint);
+#ifdef CZMQ_BUILD_DRAFT_API
+        const char *public_key = zmsg_popstr (msg);
+        server_connect (self, endpoint, public_key);
+#else
         server_connect (self, endpoint);
```
bluca: keeping server_connect the same, here just pass NULL in this branch of the ifdef, so the code should be simpler

src/zgossip.c
```diff
@@ -520,6 +555,65 @@ zgossip_test (bool verbose)
     zactor_destroy (&alpha);
     zactor_destroy (&beta);
 
+#ifdef CZMQ_BUILD_DRAFT_API
+    // curve
+    if (zsys_has_curve()) {
+        printf("testing CURVE support");
+        zclock_sleep (2000);
```
bluca: why the 2 seconds sleep before the tests? did you have any issues?

src/zgossip.c
wesyoung: ack

src/zgossip.c
wesyoung: ack, ty for this- my C is not so .. sharp :) appreciate it.

src/zgossip.c
wesyoung: most of this dev took place on vbox (vagrant, ubuntu, etc). i've noticed a few issues with doing this on a vm over the years where some race like issues crop up and timers seem to make things happy again. i'll retry (now that i know it works end to end locally) and see if it's still needed. sometimes the crypto needed time to settle too [again, on the vm in  some odd cases].

src/zgossip.c
wesyoung: err you mean straight up this:

```
if (streq (method, "CONNECT")) {
        char *endpoint = zmsg_popstr (msg);
        assert (endpoint);
        const char *public_key = zmsg_popstr (msg);
        server_connect (self, endpoint, public_key);
        zstr_free (&endpoint);
    }
```

(i didn't wanna colossally mess up someone's system while we were hammering this idea out, so everything defaulted to verbose-seperate-out, figured this would re-factor out as it was tested and  we became comfortable with it..)

bluca: Yeah that should work even better - there shouldn't be a third string in the CONNECT message so popstr will return NULL

src/zgossip.c
bluca: Yeah I've seen a few as well, if you check some tests like zproxy do extra sleeps in the CURVE case too

wesyoung:
@bluca i think there are two issues-

1. didn't think about zmq2-3 support, so i prob need to wrap some of this in 'has-curve' (or wva that function was)
2. travis trying to test beacon over tcp/ip

```
...
 * zbeacon: D: (czmq_selftest) 17-09-11 13:43:22 zsys_udp_send: failed, reason=Operation not permitted
D: (czmq_selftest) 17-09-11 13:43:22 zbeacon: failed to transmit, attempting reconnection. reason=Resource temporarily unavailable
OK
 * zgossip: testing CURVE supportI: (czmq_selftest) 17-09-11 13:43:25 zauth: API command=ALLOW
I: (czmq_selftest) 17-09-11 13:43:25 zauth: - whitelisting ipaddress=127.0.0.1
I: (czmq_selftest) 17-09-11 13:43:25 zauth: API command=CURVE
D: (czmq_selftest) 17-09-11 13:43:25 server:     API command=SET PUBLICKEY
D: (czmq_selftest) 17-09-11 13:43:25 server:     API command=SET SECRETKEY
D: (czmq_selftest) 17-09-11 13:43:25 server:     API command=AUTH ZAP DOMAIN
I: (czmq_selftest) 17-09-11 13:43:25 auth zap domain: global
D: (czmq_selftest) 17-09-11 13:43:25 server:     API command=BIND
I: (czmq_selftest) 17-09-11 13:43:25 global
D: (czmq_selftest) 17-09-11 13:43:26 client:     API command=SET PUBLICKEY
make[2]: *** [check-local] Segmentation fault (core dumped)
make[2]: Leaving directory `/home/travis/build/zeromq/czmq'
make[1]: *** [check-am] Error 2
make[1]: Leaving directory `/home/travis/build/zeromq/czmq'
make: *** [check-recursive] Error 1
```

think i can try to simplify this test a bit- over ipc, but- have to think a bit more about this. kinda want it to test this over ip but maybe i'm overthinking it for this and it doesn't really matter (i don't wanna muck up the testing for something like this..).

thoughts?

bluca:
1) Yes, check runtime cap with ```zsys_has_curve()```
2) src/zauth.c also uses TCP, check how it uses it and it should work in the same way (eg: don't hard-code the port, use a wildcard, etc etc)

wesyoung:
i mis-read what was failing. it took me a bit to reproduce it consistently.. letting travis do it's thing and will try to clear them up.. will ping you again when it's ready to go.

wesyoung:
@bluca i think i nailed down the errors i introduced (seg faults, valgrind, etc). the zmq2 test seems to fail randomly (which i don't think is because of my PR.. i saw it happening in other PRs too), the other two fails are zloop issues unrelated to this PR.

i also sync'd the branch with the overnight PRs from master.. if you think this is OK it can be merged. once it's built i can re-test the zyre PR with the new master [once it's built].

bluca:
Great, thanks, merged!

src/zgossip_engine.inc
```diff
@@ -83,6 +83,10 @@ typedef struct {
     size_t timeout;             //  Default client expiry timeout
     bool verbose;               //  Verbose logging enabled?
     char *log_prefix;           //  Default log prefix
+
```
bluca: @wesyoung there's also the problem that all these changes are going to get lost when zgossip gets regenerated, as it's a zproto-generated engine. We can keep an eye on it and keep it there manually, but long term we should think of a proper solution

wesyoung: bah- i've just started wrapping my head around zproto- that was an oversight. for some reason i was thinking the .inc wasn't auto-generated (i forget why), should have known better.

let me look through it a bit see what i can come up with. it doesn't make sense to keep it there manually (just creates confusion), if i can't find a decent path forward in the next few days- we can revert it and go from there.. thanks for catching this...

bluca: No worries, and take your time. We don't regenerate that often (quite rarely actually) and the gossip message protocol is half broken anyway when regenerated.

So it's fine to keep it as-is for now

PR #1759, opened 2017-09-12
Problems: zgossip too chatty, binds to hard-coded TCP port, out of date with zproto

bluca:
Solutions: see commits

### DIFF
--- a/include/zgossip.h
+++ b/include/zgossip.h
@@ -10,13 +10,13 @@
      * The XML model used for this code generation: zgossip.xml, or
      * The code generation script that built this file: zproto_server_c
     ************************************************************************
-    Copyright (c) the Contributors as noted in the AUTHORS file.       
-    This file is part of CZMQ, the high-level C binding for 0MQ:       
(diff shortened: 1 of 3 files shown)


wesyoung:
funny, i was just running across some of that in zyre too. forgot about the if/verbose stuff.

jimklimov:
Just in case, why change back from `freen` to `free`? 

And what's about `...get()` vs `...resolv()`?

bluca:
> Just in case, why change back from freen to free?

That's a manual change that is not in zproto so it gets overwritten - it's not very useful in these cases so it's not worth keeping it manually in place

> And what's about ...get() vs ...resolv()?

zconfig_resolve doesn't exist:

https://github.com/zeromq/czmq/blob/master/api/zconfig.api

bluca:
> zconfig_resolve doesn't exist:
>
> https://github.com/zeromq/czmq/blob/master/api/zconfig.api

Was renamed by https://github.com/zeromq/czmq/pull/988

PR #1769, opened 2017-10-11
Bugfix in nodejs binding

bjornwennberg71:
When you send a string using js that is longer than 354 bytes long (or thereabout) only garbage is sent.

Below is a patch that fixes up all the bindings thats affected.

Utf8String has 2 buffers. One is 1024 bytes long, and if the input-string exceeds this amount,
it will dynamically allocate a longer string. And it will deallocate this string in its dtor.
So, when the bindings uses the Ut8String it goes out of scope, crap happens.

I patched up all occurences of Utf8String in bindings.cc

BUGGY METHOD:
NAN_METHOD (Zstr::_send){
    Zsock *dest = Nan::ObjectWrap::Unwrap<Zsock>(info [0].As<Object>());
    char *string;
    if (info [1]->IsUndefined ())
        return Nan::ThrowTypeError ("method requires a `string`");
    else  if (!info [1]->IsString ())
        return Nan::ThrowTypeError ("`string` must be a string");
    else {
        Nan::Utf8String string_utf8 (info [1].As<String>());
        string = *string_utf8;
	// BUG HAPPENS HERE
	// Utf8String::dtor() happens and our 'string' points to crap
    }
    int result = zstr_send (dest->self, (const char *)string);
    info.GetReturnValue ().Set (Nan::New<Number>(result));

}

CORRECT METHOD:
NAN_METHOD (Zstr::_send){
    Zsock *dest = Nan::ObjectWrap::Unwrap<Zsock>(info [0].As<Object>());
    char *string;
    if (info [1]->IsUndefined ())
        return Nan::ThrowTypeError ("method requires a `string`");
    else  if (!info [1]->IsString ())
        return Nan::ThrowTypeError ("`string` must be a string");
    else {
        Nan::Utf8String string_utf8 (info [1].As<String>());
        string = *string_utf8;

        int result = zstr_send (dest->self, (const char *)string);
        info.GetReturnValue ().Set (Nan::New<Number>(result));
    }

}

# Pull Request Notice

```
Problem: X is broken

Solution: do Y and Z to fix X
```

Please avoid sending a pull request with recursive merge nodes, as they
are impossible to fix once merged. Please rebase your branch on
zeromq/czmq master instead of merging it.

git remote add upstream git@github.com:zeromq/czmq.git
git fetch upstream
git rebase upstream/master
git push -f

In case you already merged instead of rebasing you can drop the merge commit.

git rebase -i HEAD~10

Now, find your merge commit and mark it as drop and save. Finally rebase!

If you are a new contributor please have a look at our contributing guidelines:
[CONTRIBUTING.md](https://github.com/zeromq/czmq/blob/master/CONTRIBUTING.md)

### DIFF
--- a/bindings/nodejs/binding.cc
+++ b/bindings/nodejs/binding.cc
@@ -81,11 +81,12 @@ NAN_METHOD (Zargs::New) {
     else {
         Nan::Utf8String argv_utf8 (info [1].As<String>());
         argv = *argv_utf8;
-    }
-    Zargs *zargs = new Zargs ((int) argc, (char **)&argv);
-    if (zargs) {
-        zargs->Wrap (info.This ());
-        info.GetReturnValue ().Set (info.This ());
+    
+        Zargs *zargs = new Zargs ((int) argc, (char **)&argv);
+        if (zargs) {
+          zargs->Wrap (info.This ());
+          info.GetReturnValue ().Set (info.This ());
+        }
     }
 }
 
@@ -153,9 +154,10 @@ NAN_METHOD (Zargs::_param_lookup) {
     else {
         Nan::Utf8String keys_utf8 (info [0].As<String>());
         keys = *keys_utf8;
+    
+        char *result = (char *) zargs_param_lookup (zargs->self, (const char *)keys);
+        info.GetReturnValue ().Set (Nan::New (result).ToLocalChecked ());
     }
-    char *result = (char *) zargs_param_lookup (zargs->self, (const char *)keys);
-    info.GetReturnValue ().Set (Nan::New (result).ToLocalChecked ());
 }
 
 NAN_METHOD (Zargs::_param_lookupx) {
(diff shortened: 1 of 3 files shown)


bluca:
Hi, thanks for the PR, but these bindings are autogenerated. Could you please fix them in https://github.com/zeromq/zproject/blob/master/zproject_nodejs.gsl ? Otherwise the next time the regeneration happen all your changes will be lost.

bjornwennberg71:
Sure thing. I also added two scripts to test to my branch of czmq

Thanx
Bjornw

11. okt. 2017 16:45 skrev "Luca Boccassi" <notifications@github.com>:

vyskocilm:
Hi, this PR is 14 days old - I'll close this one and will wait for new one.

@bjornwennberg71 please change this file ... https://github.com/zeromq/zproject/blob/master/zproject_nodejs.gsl and run make code in czmq.

PR #1775, opened 2017-10-25
Problem: zproject's selftest-rw does not allways work

vyskocilm:
... we run our tests inside complicated Docker containers setup, where is no
guarantee user running test can access files in src/selftest-rw.

Solution: Add a constructor to create temporary file, which is automatically
removed on destroy. This will leads to nicer API for tests + it'd be hard to
find a platform without writable temporary directory. So this concept can work
better than src/selftest-rw + it leads to nicer code and simpler Makefiles generated by zproject.

XXX: it does use tmpnam, because I did not find a portable way how to obtain filename from FILE* or fd. On non-Windows the problem is mitigated by O_CREAT|O_EXCL flags for open, which will fail if file exists.

### DIFF
--- a/api/zfile.api
+++ b/api/zfile.api
@@ -20,6 +20,11 @@
         <argument name = "name" type = "string" />
     </constructor>
 
+    <constructor name="tmp">
+        Create new temporary file for writing via tmpfile. File is automaticaly
+        deleted on destroy
+    </constructor>
+
     <destructor>
         Destroy a file item
     </destructor>
--- a/api/zfile.api
+++ b/api/zfile.api
@@ -20,7 +20,7 @@
         <argument name = "name" type = "string" />
     </constructor>
 
-    <constructor name="tmp">
+    <constructor name="tmp" state="draft">
         Create new temporary file for writing via tmpfile. File is automaticaly
         deleted on destroy
     </constructor>
--- a/src/zfile.c
+++ b/src/zfile.c
@@ -890,6 +890,7 @@ zfile_test (bool verbose)
     zfile_close (file);
     zfile_destroy (&file);
 
+#ifdef CZMQ_BUILD_DRAFT_API
     zfile_t *tempfile = zfile_tmp ();
     assert (tempfile);
     assert (zfile_filename (tempfile, NULL));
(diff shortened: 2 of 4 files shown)


api/zfile.api
```diff
@@ -20,6 +20,11 @@
         <argument name = "name" type = "string" />
     </constructor>
 
+    <constructor name="tmp">
```
bluca: Could you please mark it as draft given it's new?

vyskocilm: Done

src/zfile.c
```diff
@@ -100,9 +106,59 @@ zfile_new (const char *path, const char *name)
...
+    assert (self);
+    self->remove_on_destroy = true;
+
+    // I know tmpnam is considered insecure, however I did not find a way
+    // how to get fullname from FILE *, or fd, which would be
+    // portable. If someone knows the way, feel free to rewrite it.
+    // ... on non Windows platforms there is O_CREAT | O_EXCL for open,
+    // which prevents the temp file name attacks
+    char name [L_tmpnam];
+    char *r = tmpnam (name);
```
bluca: Unfortunately this breaks the build on OSX:

```
src/zfile.c:133:15: error: 'tmpnam' is deprecated: This function is provided for compatibility reasons only. Due to security concerns inherent in the design of tmpnam(3), it is highly recommended that you use mkstemp(3) instead. [-Werror,-Wdeprecated-declarations]
 char *r = tmpnam (name);
```

https://travis-ci.org/zeromq/czmq/jobs/292505743

In libzmq we use mkdtemp or mkstemp if that's not available:

https://github.com/zeromq/libzmq/blob/master/src/ipc_listener.cpp#L105

api/zfile.api
bluca: great, thanks

src/zfile.c
bluca: An alternative could be to disable that warning with a pragma, at least temporarily

vyskocilm: It that would be for Linux, it'd be easy. However right now I don't see any easy way ...

Plan to do
* on windows I will keep it unimplemented atm
* if we have F_GETPATH we can use tmpfile, which have almost the same semantics
* if we don't (Linux) then use mkstemp + proc reading

vyskocilm:
Due issues with OSX I am closing the PR, needs to be reworked

jimklimov:
Hi. Just in case - the user can build the program but can not write into that src dir? Or build is completely separate from test? Is there SOME location that the user running the test can write into (e.g. `/tmp`)?

In the latter case, you can revert the recent removal of variables (rather than macros) used for selftest dir names, and implement the ability to provide envvar setting of those variables to override the default of using the build directory - which is simple but we never got around to do, as our project's practice did not need that bit.

PR #1778, opened 2017-10-31
Problem: zfile_tmp does not work on Windows

vyskocilm:
Solution: hack something using MSDN as a guide

@bluca this is mainly to test the MSDN copy, I don't have any Windows there, so can't really test it. This way it will be built by AppVeyor.


vyskocilm:
It looks that Windows path works. Do not understand the Travis Failure - I see error in generating build script only.

bluca:
Yeah that was weird, I've kicked it let's see if it starts now

PR #1779, opened 2017-10-31
Problem: zproxy_test does construct variables the hardest way

vyskocilm:
Solution:
 1.) Move them to appropriate place in the test
 2.) It does not make a sense to cat two constant strings using
     sprintf. Simply declare them with a content they should have.

@bluca I'd like to trigger Travis on an another PR.

### DIFF
--- a/src/zproxy.c
+++ b/src/zproxy.c
@@ -484,23 +484,6 @@ zproxy_test (bool verbose)
         printf ("\n");
 
     //  @selftest
-
-    const char *SELFTEST_DIR_RW = "src/selftest-rw";
-
-    const char *testbasedir  = ".test_zproxy";
-    const char *testpassfile = "password-file";
-    const char *testcertfile = "mycert.txt";
-    char *basedirpath = NULL;   // subdir in a test, under SELFTEST_DIR_RW
-    char *passfilepath = NULL;  // pathname to testfile in a test, in dirpath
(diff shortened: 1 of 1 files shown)


bluca:
This one was fine, no idea what's wrong with the other, kicked it again

jimklimov:
IMHO this slew of Michal's fixes back from the selftest support is a regression in regard to coding style.

I sort of agree that the pre-generated code is more complex and clunky than it could be (and sorry for not recognizing how it leaked into generated docs as well), but it also provides(-ed?) a template that developers could just copy-paste and use in their selftests, as well as a sort of reference implementation of such tests for them to consult. Now different parts of the same codebase are mixed-style again.

Note that the predefined values and concatenation aimed for a couple of goals:
* easier modification, so if selftest dirs were to get set via envvars for example (e.g. to discover discrepancies of VFAT/NTFS/ext4/... timestamp handling where it matters), the change is in one place of the sources, and effect is at run-time;
* readability, in way of seeing what is where and why by construction of the variable

When I revised and changed tests to this pattern, it was not all for the fun of doing so, but because all this combing through the selftest code fixed actual bugs and CI failures - similar to how others fixed network-related tests from hardcoded ports to dynamically allocated ones - and made it more understandable to read. Therefore, going back to hardcoded ways is a regression... at least, so far the variables with paths are still there. Next logical step should be to remove them too and hardcode paths back into each command again?

PR #1782, opened 2017-11-02
PROBLEM: Can't remove zconfig subtree

karolhrdina:
SOLUTION: Implement it; Write a test.

### DIFF
--- a/api/zconfig.api
+++ b/api/zconfig.api
@@ -189,7 +189,7 @@
         <argument name = "file" type = "FILE" />
     </method>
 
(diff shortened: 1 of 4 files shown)


api/zconfig.api
```diff
@@ -180,6 +180,10 @@
         <return type = "boolean" />
     </method>
 
+    <method name = "remove">
```
vyskocilm: Please add state="draft"

karolhrdina: Done.
Thanks for the notice.

Issue #1785, opened 2017-11-10
zfile.c -- E0167 argument of type "TCHAR *" is incompatible with parameter of type "const char *"


sgluss:
this error is occurring in Windows 10, Visual Studio 2017 targeting x64 with platform toolset 141

error occurs at line 147 in zfile.c:
`self->fullname = strdup (szTempFileName);`

I was able to hack-fix it by adding a cast to szTempFileName:
`self->fullname = strdup ((char*)szTempFileName);`

this enables the build to succeed, but is definitely a banana-in-the-differential type fix. What do you guys think?

vyskocilm:
Hi,

Does make check work?  I don't know a lot about windows programming, but it
looks fine to me. Can you send the pull request?

Dne 10. 11. 2017 6:37 odpoledne napsal uživatel "Samuel Gluss" <
notifications@github.com>:

twhittock:
Perhaps you have somehow enabled wide-char support with the UNICODE define or in the project settings, and converted the standard library calls to wide-char versions. If so, the cast would cause issues.

sgluss:
@vyskocilm : ran the tests, all of them pass:
`
C:\czmq\czmq\builds\msvc\vs2015\x64\DebugDEXE>czmq_selftest.exe
Running czmq selftests...
...
Tests passed OK
`
I'll prepare a PR

@twhittock I already experimented a bit with the project's char settings, I'll try experimenting with it some more just be be sure I was thorough

sgluss:
I tried adjusting the `Character Set` settings in the libczmq project to no avail... without the cast, I get the error no matter what I set. Note the inherited preprocessor settings though, there's unicode right there! 
![libczmqunicodeinpreprocessor](https://user-images.githubusercontent.com/12822580/32692448-1909c586-c6cd-11e7-84cb-0c5cbb3ccc5d.png)
I'll keep digging and see if this is the cause of the problem, any other insights would be great!

evoskuil:
TCHAR macros should not be used in a library that works almost with char. I proposed converting the library to UTF8 Everywhere, but that is a large task. Casting is not a solution.

AntonBarwald:
@evoskuil I guess this is not a "solution/workaround" either? 
```
wcstombs(self->fullname, szTempFileName, MAX_PATH);
```

evoskuil:
[...]

Issue #994, opened 2015-04-21
Problem: Unicode paths, etc. not supported in Windows.


evoskuil:
This is a pervasive problem. For example:

``` c
int
zsys_file_delete (const char *filename)
{
    assert (filename);
#if (defined (__WINDOWS__))
    return DeleteFileA (filename)? 0: -1;
#else
    return unlink (filename);
#endif
}
```

In this example (and all others) a non-ASCII path will only be properly interpreted if all of the characters in the path are members of the current thread's ANSI code page. This may or may not be the case, but in any case there is no way to support characters that span multiple code pages or that don't exist in any.

In most Windows APIs the only way to parameterize the call with Unicode input is to pass `wchar_t` (wide) parameters. For example:

``` c
int
zsys_file_delete (const char *filename)
{
    assert (filename);
#if (defined (__WINDOWS__))
    /* insert code to convert filename to wide_filename here */
    return DeleteFileW (wide_filename)? 0: -1;
#else
    return unlink (filename);
#endif
}
```

However the conversion from `char` to `wchar_t` requires code page mapping. The only solution is to carry the Unicode text throughout the entire code path. Traditionally this has been done in in VC projects by using `wchar_t` everywhere. That of course results in a pervasive conditional string/char.

Code like this has the same problem in reverse (mbcs is ANSI, not Unicode or ASCII):

``` c
    ...
    PWCHAR friendlyName = cur_address->FriendlyName;
    size_t asciiSize = wcstombs (0, friendlyName, 0) + 1;
    char *asciiFriendlyName = (char *) zmalloc (asciiSize);
    wcstombs (asciiFriendlyName, friendlyName, asciiSize);
```

The better approach is to use [utf8 everywhere](http://utf8everywhere.org). This works well and is not hard or disruptive. I've recently implemented the technique across all libbitcoin repos, including two console applications, with good results. It would take some time, and probably be best to start by providing the tools in `libzmq`.

I've hit this issue myself in working with `zcert` paths. That can probably be mitigated by providing a round-trip stream alternative, but it would be best for ZMQ to resolve this universally. I'm interested to hear thoughts on the idea.

hintjens:
I'm absolutely in favor of banning UTF16 and using UTF8 everywhere.

evoskuil:
As a general rule I would wrap such calls with unpublished utility functions to hide the Unicode conversions. These would use published UTF8<->UTF16 conversion utilities that need only be exposed for Windows, built on the native conversion APIs.

[...]

PR #1786, opened 2017-11-11
added cast to correct error: "TCHAR *" is incompatible with parameter of type "const char *"

sgluss:
see issue and discussion of fix here:
https://github.com/zeromq/czmq/issues/1785


bluca:
@evoskuil what else could be done to fix the problem? Are you familiar with the Windows API being used there?

evoskuil:
I am very familiar but unfortunately traveling with only my phone, so hard to do much. Ultimately UTF8 Everywhere is the proper solution. Few projects have done this, but we have done so in Libbitcoin. See other thread for quick fix recommendations.

bluca:
@evoskuil ok thanks. Given there's a build breakage and it's a new draft API so not very used, I propose to merge this as a quick bandaid to get the whole library working again in Windows, and then a proper fix can be researched without rush.

sgluss:
In the meantime I'll keep investigating why VS2017 is having trouble with the TCHAR. This code compiled without complaint in Ubuntu 16.04

evoskuil:
[Issue 994](https://github.com/zeromq/czmq/issues/994) explains the core issue in more detail.

evoskuil:
This is a local fix: https://github.com/zeromq/czmq/pull/1793

PR #1787, opened 2017-11-12
Problem: Homebrew must run under Ruby 2.3

AntonBarwald:
Solution: Run 'brew update' before 'brew install package_name'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,7 @@ addons:
     - libzmq3-dev
 
 before_install:
+- if [ $TRAVIS_OS_NAME == "osx" ]; then brew update ; fi
 - if [ $TRAVIS_OS_NAME == "osx" -a $BUILD_TYPE == "android" ] ; then brew install binutils ; fi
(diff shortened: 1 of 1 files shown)


AntonBarwald:
Maybe there is a drawback to do 'brew update' each time?

bluca:
This should be solved when a new osx image is deployed on Wednesday, that's why I've left it alone. A brew update unfortunately takes minutes, making the already slow osx builds even slower :-(

AntonBarwald:
@bluca Yes, I saw that it was making it slower 👎. Should I close this one since it will be fixed with new OS X image on Wednesday?

bluca:
Yes please, if you don't mind. We can then reopen if the deployment doesn't solve it: https://blog.travis-ci.com/2017-10-16-a-new-default-os-x-image-is-coming

bluca:
Looks like they haven't fixed it yet. Slightly different fix to run brew update only in builds that actually need brew https://github.com/zeromq/czmq/pull/1794

PR #1793, opened 2017-11-14
Problem: inconsistent use of TCHAR will break non-unicode VC++.

evoskuil:
Solution: use only explicitly ANSI Windows APIs. Also fix style errors.

There is no point in using Unicode or conditionally-Unicode (TCHAR) APIs in most cases as zeromq does not utilize wchar_t or TCHAR string types. In order to be Unicode capable in Windows zeromq should implement [UTF8 Everywhere](http://utf8everywhere.org).


bluca:
Thanks!

PR #1798, opened 2017-11-26
Fix autotools build on Cygwin

gitplcc:
Fix autotools build on Cygwin failing due to:

- O_BINARY symbol redefinition in include/czmq_prelude.h
- __attribute__ ((visibility ("hidden"))) not supported by Windows executables.


bluca:
Thanks but as the header says the first file is generated so your change will be overwritten, you need to fix it in https://github.com/zeromq/zproject

PR #1809, opened 2017-12-07
Problem: Jenkinsfile does not pass distcheck etc.

jimklimov:
Solution: tune the recipe so it passes... renamed test dir (unstash in another agent) complicates things for libtool... Hope to hack a cleaner solution, ultimately generatable with zproject, at some point. But so far a more important value would be for this to just work - even if suboptimally.


jimklimov:
Results are random-failing at zproc calling `src/zsp` and sometimes grabbing no stdout. Sometimes it passes all *check subtests, sometimes it fails one - and that's it :(

Issue #1811, opened 2017-12-11
Not able to configure static library build


glaure:
It would be nice to be able to configure static and shared library builds.
Similar to zmq's CMakeLists, there should be libczmq.so and libczmq-static.a.

If somebody could hint me how to use zproject, I could try it to solve this.

Currently I am editing CMakeLists.txt manually and in fear of the files beginning line:
################################################################################
#  THIS FILE IS 100% GENERATED BY ZPROJECT; DO NOT EDIT EXCEPT EXPERIMENTALLY  #
#  Read the zproject/README.md for information about making permanent changes. #
################################################################################

 :-)

Bye Gunther

bluca:
https://github.com/zeromq/zproject

edit the zproject_cmake.gsl file to change the template for cmake - you'll need the gsl program to regenerate, from https://github.com/zeromq/gsl

glaure:
Thanks a lot! I will try my best

bluca:
Fixed by #1813

PR #1813, opened 2017-12-12
New CMakeLists.txt with static/shared library support generated with zproject

glaure:
Problem: Not able to configure static library build
https://github.com/zeromq/czmq/issues/1811

Solution: 
Changes to zproject allow to configure static and shared library builds.
This commit includes a new generated CMakeLists.txt file.

Tested on Linux and Windows.


bluca:
I think this broke appveyor's tests

glaure:
How can I help? I have no experience with appveyor.

bluca:
Changing the filename is backward incompatible as it breaks existing scripts, so yes please restore the previous name, thanks

Issue #1401, opened 2016-04-15
README.md / Building on Windows needs to be updated


vtellier:
Hi,

While trying to compile on Windows I followed the steps of the README.md, but I got stucked.

```
git clone git://github.com/zeromq/czmq.git
cd czmq\builds\msvc
configure.bat
cd build
buildall.bat
```

The folder build doesn\t exists anymore since the commit hintjens@ef4386fd8d1e4dc8b315640099748d2f2dc23145.

> This tree contains a mix of zproject generated files, MSVC produced
> files, and hand-made files. This is unmaintainable.
> 
> Solution:
> - ignore ALL MSVC generated files in .gitignore
> - generate all necessary files in zproject (such as build.bat)
> - delete the rest here

It would be nice to more details about that in README.md

ahmetkakici:
Hello @vtellier,
You don't need build folder, just change directory to the version you want to build for.
For example, you can execute below steps to build for vs2013.


```
git clone git://github.com/zeromq/czmq.git
cd czmq\builds\msvc
configure.bat
cd vs2013
build.bat
```

sappo:
@vtellier does it work the same for libzmq?

sappo:
Sorry @vtellier I meant to address @ahmetkakici.

ahmetkakici:
@sappo i've build both libzmq and czmq vs2013 solutions on my windows machine, what is the problem you've encountered?

sappo:
@ahmetkakici no problem! I'd liked to confirm the build process though before updating the README.

ahmetkakici:
@sappo sorry i've been misunderstood. 
libzmq still have build directory and build.bat file, but configre.bat removed from libzmq. 
Let me build both libzmq, libsodium and czmq from scratch and check readme file.

sappo:
@ahmetkakici great thanks!

ahmetkakici:
With https://github.com/zeromq/czmq/pull/1468 I've added buildall.bat file to czmq repository.
No need for updating readme file.

xoorath:
Hey, making note over a year later on this issue... the Readme is also inherently fragile by just suggesting users clone the latest commit from dependent repos.  

Maybe the readme could be updated to:
1) work correctly with the current state of dependent projects, and vs2017.
2) tag specific releases so if it gets run some number of years later (such as my case), it will still work correctly, even if not updated regularly.

[...]

PR #1815, opened 2017-12-12
Updating readme for issue #1401

xoorath:
Problem: readme.md gives fragile (and currently broken) instructions for building on windows.

Solution: remove until someone who knows how to build for windows reliably has time to update the readme, or someone who has times finds out how to build reliably; as not to leave new users hanging.

See #1401

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 
 **[Building and Installing](#building-and-installing)**
 
-**[Building on Windows](#building-on-windows)**
+**[~Building on Windows~](#building-on-windows)**
 
 **[Linking with an Application](#linking-with-an-application)**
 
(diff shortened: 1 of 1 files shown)


bluca:
These instructions were added by someone for whom they were working. Windows is notoriously quirky and easily broken, so perhaps there's something wrong with your environment. If you want to amend or fixup it's fine, but removing completely is not, as something that's not working for you might very well work for somebody else.

xoorath:
@bluca Yes, it's completely possible there's something on my workstation preventing me from building. I haven't figured that out yet.

However considering the absence of `cmzq/builds/msvc/buildall.bat` referenced in the `readme.md`, and the inherit fragility of suggesting users clone the root of 3 repos and hope they work together on the day and hour they type clone: I would suspect pretty highly that these instructions are broken despite the nature of developing for windows.

PR #1817, opened 2017-12-13
Removed versioned library naming

glaure:
Because of compatibility issues, versioned DLL library naming had to be removed.
Hopefully appveyor tests succeed again.

Tested on Linux and Windows.


bluca:
Thanks! The tests are passing again now

PR #1819, opened 2017-12-19
PROBLEM: Need to remove any node and subtree

karolhrdina:
not just root.
    
Previously a method `zconfig_remove` had been created that removes just
the subtree (all children) of the selected node, which is confusing.

SOLUTION:

* Rename `zconfig_remove` -> `zconfig_remove_subtree`
* Use name `zconfig_remove` for the _new_ method
* Implement the new method
* Write test for `zconfig_remove`; expand test for `zconfig_remove_subtree`


karolhrdina:
There are merge commits in my PR.

PR #1820, opened 2017-12-19
PROBLEM: Need to remove any node and subtree

karolhrdina:
not just root.
    
Previously a method `zconfig_remove` had been created that removes just
the subtree (all children) of the selected node, which is confusing.
    
SOLUTION:

* Rename `zconfig_remove ()` -> `zconfig_remove_subtree ()`
* Use name `zconfig_remove` for the _new_ method
* Implement `zconfig_remove ()`
* Write test for `zconfig_remove ()`; Extend test for `zconfig_remove_subtree ()`

### DIFF
--- a/api/zconfig.api
+++ b/api/zconfig.api
@@ -180,8 +180,13 @@
         <return type = "boolean" />
     </method>
 
-    <method name = "remove" state = "draft">
-        Destroy subtree (child)
+    <method name = "remove_subtree" state = "draft">
+        Destroy subtree (all children)
+    </method>
+
+    <method name = "remove" singleton = "1" state = "draft">
+        Destroy node and subtree (all children)
+        <argument name = "self_p" type = "zconfig" by_reference = "1" />
     </method>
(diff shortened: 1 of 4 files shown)


bluca:
@karolhrdina @sappo JNI build fails due to the new zconfig_remove, the generated code is apparently wrong:

```
/home/travis/build/bluca/czmq/bindings/jni/src/main/java/org/zeromq/czmq/Zconfig.java:236: error: incompatible types: unexpected return value
        return 0;
               ^
1 error
1 warning
 FAILED
FAILURE: Build failed with an exception.
```

I've fixed it manually for now:

https://github.com/zeromq/czmq/pull/1822/commits/44082761b1d86ace3a25b747333eba67d860e721

Any idea what the problem could be?

sappo:
I'll have a look at it.

Am 25.12.2017 7:19 nachm. schrieb "Luca Boccassi" <notifications@github.com

PR #1829, opened 2018-01-03
Fix zuuid.c build error on FreeBSD

jj1bdx:
Problem: zuuid.c on v4.1.0 is broken on FreeBSD 11.1-STABLE

Solution: apply fixes in this pull request for include/czmq_prelude.h and src/zuuid.c

* This fix works on FreeBSD 10.4-RELEASE and 11.1-RELEASE, 11.1-STABLE
* Forcefully define HAVE_UUID in czmq_prelude.h if FreeBSD
    - FreeBSD has the stock UUID library, but it's not -luuid, it's in -lc, so HAVE_UUID is not defined during configure
* Skip code in zuuid.c assuming uuid_generate(2) if FreeBSD
    - FreeBSD does not have uuid_generate(2) (it's Linux specific); use existing code for FreeBSD instead

See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=224853 for the details

### DIFF
--- a/src/zuuid.c
+++ b/src/zuuid.c
@@ -47,7 +47,7 @@ zuuid_new (void)
     assert (sizeof (uuid) == ZUUID_LEN);
     UuidCreate (&uuid);
     zuuid_set (self, (byte *) &uuid);
-#elif defined (HAVE_UUID)
+#elif defined (HAVE_UUID) && !defined (__UTYPE_FREEBSD)
     uuid_t uuid;
     assert (sizeof (uuid) == ZUUID_LEN);
     uuid_generate (uuid);
--- a/include/czmq_prelude.h
+++ b/include/czmq_prelude.h
@@ -683,6 +683,9 @@ char *if_indextoname (unsigned int ifindex, char *ifname);
 #if defined (__UTYPE_OSX) && !defined (HAVE_UUID)
 #   define HAVE_UUID 1
 #endif
+#if defined (__UTYPE_FREEBSD) && !defined (HAVE_UUID)
+#   define HAVE_UUID 1
+#endif
 #if defined (HAVE_UUID)
 #   if defined (__UTYPE_FREEBSD) || defined (__UTYPE_NETBSD) || defined(__UTYPE_OPENBSD)
 #       include <uuid.h>


bluca:
Thanks! Have a feeling we already fixed something like this...

src/zuuid.c
```diff
@@ -47,7 +47,7 @@ zuuid_new (void)
     assert (sizeof (uuid) == ZUUID_LEN);
     UuidCreate (&uuid);
     zuuid_set (self, (byte *) &uuid);
-#elif defined (HAVE_UUID)
+#elif defined (HAVE_UUID) && !defined (__UTYPE_FREEBSD)
```
bluca: @jj1bdx looking again at this as there's a build failure in debian-kfreebsd (debian userland + freebsd kernel):

```
src/zuuid.c: In function 'zuuid_new':
src/zuuid.c:58:5: error: implicit declaration of function 'uuid_create'; did you mean 'uuid_clear'? [-Werror=implicit-function-declaration]
     uuid_create (&uuid, &status);
     ^~~~~~~~~~~
     uuid_clear
src/zuuid.c:59:19: error: 'uuid_s_ok' undeclared (first use in this function); did you mean 'zuuid_str'?
     if (status != uuid_s_ok) {
                   ^~~~~~~~~
                   zuuid_str
src/zuuid.c:59:19: note: each undeclared identifier is reported only once for each function it appears in
src/zuuid.c:64:5: error: implicit declaration of function 'uuid_enc_be'; did you mean 'uuid_type'? [-Werror=implicit-function-declaration]
     uuid_enc_be (&buffer, &uuid);
     ^~~~~~~~~~~
     uuid_type
```

https://buildd.debian.org/status/fetch.php?pkg=czmq&arch=kfreebsd-amd64&ver=4.1.1-1&stamp=1522679203&raw=0

There's something I don't understand - this condition here skips this branch if HAVE_UUID is set but it's building on FreeBSD, but HAVE_UUID is set on FreeBSD only because the change in this patch added it? So why was this change necessary at all?

jj1bdx: I'm sorry I have no testing environment of debian-kfreebsd here. I'll repeat my reason:

> FreeBSD has the stock UUID library, but it's not -luuid, it's in -lc, so HAVE_UUID is not defined during configure

I have no idea whether on debian-kfreebsd the situation is the same. I'm running this on FreeBSD amd64 11.1-STABLE r331481.

bluca: I read that, what I can't grok is why doing:
```
#define HAVE_UUID // this is manually defined
...
#elif defined (HAVE_UUID) && !defined (__UTYPE_FREEBSD) // now this has to be manually skipped
...
#elif defined (__UTYPE_FREEBSD) // to get here
```

rather than just leaving

```
#(un)define HAVE_UUID // leave this undefined as it is by default
...
#elif defined (HAVE_UUID) // this is now skipped since HAVE_UUID is not set
...
#elif defined (__UTYPE_FREEBSD) //falls directly to the intended branch
```

?

jj1bdx: Some background reasons for my patch in zuuid.c: (Note that "FreeBSD" stands for the FreeBSD Foundation's FreeBSD, not the debian-kfreebsd)

* FreeBSD *does have* UUID so `HAVE_UUID` should be set as well as OSX or for the other OSes
* Linux and FreeBSD userlands have different library function names and implementations on uuid (as shown in the compilation error messages)

I will not object streamlining the conditional compilation clauses so long as the compilation result is unchanged.

PR #1833, opened 2018-01-19
Problem: zproc readability and usability issues

vyskocilm:
Solution: put macro magic to czmq_prelude.h, so other users will benefit from it. Simplify the zproc.c code a bit.

### DIFF
--- a/include/czmq_prelude.h
+++ b/include/czmq_prelude.h
@@ -393,6 +393,9 @@
 #   define PATH_MAX 1024
 #endif
 
+#if defined (WIN32)
+#   define getcwd _getcwd
+#endif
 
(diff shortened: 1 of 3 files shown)


jimklimov:
Good idea. Would you care to also change the 2 or so other usecases? Or are they not as simple (iirc one had different logics too)?

jimklimov:
Good idea. Would you care to also change the 2 or so other usecases? Or are they not as simple (iirc one had different logics too)?

vyskocilm:
@bluca  and otherts - sorry for pushing more patches there, I thought the first PR is already merged :-D

jimklimov:
Quite LGTM! Thanks for cleaning this up! ;)

PR #1835, opened 2018-01-22
Problem: can't build czmq on osx

vyskocilm:
Solution: use _NSGetEnviron () routine to get environ variable on OSX

I've no access to OSX, someone should check if it builds and works before merging. Should fix #1834

PR #1836, opened 2018-01-22
Problem: ergnomy of zproc API is bad

vyskocilm:
Solution: make better API, fix testing order, improve documentation. See individual commits.


vyskocilm:
Merging myself - the environ build issues are addressed in #1835

PR #1838, opened 2018-01-30
Update Jenkinsfile according to current zproject template

jimklimov:


jimklimov:
Seems something got crippled, it does not find the `zsp` in some tests ;\

PR #1841, opened 2018-02-02
Problem: zproc starting and stopping process is not implmented on win

thalman:
Solution: implement it


diorcety:
the definition of SIGKILL is in a generated header.
SIGKILL is not defined on Windows at ALL (https://msdn.microsoft.com/fr-fr/library/xdkz3x12.aspx). Should'nt we use SIGTERM on Windows ? (#define SIGTERM         15  // Software termination signal from kill)

bluca:
Could you please send a PR to change it?

diorcety:
@thalman Do you agree?

PR #1842, opened 2018-02-08
Problem: We would like to use tabs in ZPL configs.

jimklimov:
Solution: Add the support, count tab as 1 level / 4 spaces.

Signed-off-by: Jana Rapava <janarapava@eaton.com>
Forwarded-by: Jim Klimov <evgenyklimov@eaton.com>

### DIFF
--- a/src/zconfig.c
+++ b/src/zconfig.c
@@ -736,11 +736,18 @@ static int
 s_collect_level (char **start, int lineno)
 {
     char *readptr = *start;
-    while (*readptr == ' ')
+    int spaces = 0;
+    int tabs = 0;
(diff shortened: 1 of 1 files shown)


bluca:
tabs are evil :-P

karolhrdina:
I would like to point out that reason supplied in PR is not a problem.

The issue can be easily solved by configuring your favourite editor to expand to spaces when tab is pressed, which is generally recommended anyway.

This PR only complicates parsing code and provides no test.

@bluca or @sappo I would like to ask you to revert this PR, since i don't have write access.

karolhrdina:
Reverted by 
https://github.com/zeromq/czmq/pull/1843

PR #1843, opened 2018-02-08
Revert "Problem: We would like to use tabs in ZPL configs."

karolhrdina:
Reverts zeromq/czmq#1842

Two reasons:

1. PR does not solve a problem. You can set your favourite editor to expand tabs.

2. The [ZPL RFC specification](https://rfc.zeromq.org/spec:4/ZPL/) clearly states that `Hierarchy is signaled by indentation, where a child is indented *4 spaces* more than its parent.`

### DIFF
--- a/src/zconfig.c
+++ b/src/zconfig.c
@@ -736,18 +736,11 @@ static int
 s_collect_level (char **start, int lineno)
 {
     char *readptr = *start;
-    int spaces = 0;
-    int tabs = 0;
-    while (*readptr == ' ' || *readptr == '\t') {
-        if (*readptr == ' ')
-            spaces++;
-        else
-            tabs++;
+    while (*readptr == ' ')
         readptr++;
-    }
-    ptrdiff_t level = (spaces + 4*tabs) / 4;
-    if (level * 4 != spaces + 4*tabs) {
-        zclock_log ("E (zconfig): (%d) indent 4 spaces at once or use tabs", lineno);
+    ptrdiff_t level = (readptr - *start) / 4;
+    if (level * 4 != readptr - *start) {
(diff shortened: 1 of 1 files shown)


jimklimov:
Makes sense. Sorry I missed the RFC definition in my (coarse) review.

EldoreiJK:
Actually, isn't it a good time to move on and update the RFC? Given that config files are there for administrators, I hardly see any of those going through the internet to search for RFC in question when they want to configure end product.
I'm willing to update this a bit more to use "smart" indentation (find first spacing size, then use it instead of const 4), add warnings to output that config has inconsistent or wrong indentation, and actually update zconfig to be more (end) user friendly.
Please consider this as a comment given to such RFC, as being commented was original purpose of RFC. Or if you prefer to see such comment at different place, please let me know where to put it.

sappo:
Feel free to create a new RFC draft that extended the current version of
the stable RFC and send a PR.

Am 09.02.2018 9:21 vorm. schrieb "Jiri Kukacka" <notifications@github.com>:

Actually, isn't it a good time to move on and update the RFC? Given that
config files are there for administrators, I hardly see any of those going
through the internet to search for RFC in question when they want to
configure end product.
I'm willing to update this a bit more to use "smart" indentation (find
first spacing size, then use it instead of const 4), add warnings to output
that config has inconsistent or wrong indentation, and actually update
zconfig to be more (end) user friendly.
Please consider this as a comment given to such RFC, as being commented was
original purpose of RFC. Or if you prefer to see such comment at different
place, please let me know where to put it.

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/zeromq/czmq/pull/1843#issuecomment-364366271>, or mute
the thread

Issue #1844, opened 2018-02-08
make check segfaults when no USER variable is set


felixfrank:
Tried installing czmq in an archlinux docker container, using the standard:
`./autogen.sh && ./configure --prefix=... && make check`
However the tests always segfaulted.
After running make debug I found out the exact line, which is in:
https://github.com/zeromq/czmq/blob/f5475cc63f2a1544f7fc345b3ac9a6743b2d5860/src/zproxy.c#L563

printenv shows no USER variable and running the make check with a user variable makes the test succeed: 
`env USER="xyz" make check`

Maybe there's a tiny workaround for that?

bluca:
I guess it should do instead something like:

```
if (getenv("USER") && streq(getenv("USER"), "vagrant"))
```

IIRC strcmp barfs if one of the pointers is NULL which is probably what you are seeing. Could you please send a PR?

wesyoung:
bah- that's on me- should have thought about that a bit better (or at-least tested it) when i added it.

PR #1847, opened 2018-02-09
Fix segfault when no USER environment variable exists

felixfrank:
# Problem:
streq(getenv(envvar), ...) segfaults when the envvar variable does not exist because the return value of getenv is 0

# Solution:
Extend the if condition with a short-circuit check of the getenv's return value 

Fixes zeromq/czmq#1844


bluca:
Thanks!

PR #1850, opened 2018-02-11
[WIP] problem: zproject out of date

wesyoung:
i need a second pair of eyes on this- i can't remember if there are some odd fixes that still need doing (java?) or if i'm stomping on something (it's been a while), i don't see anything obvious..

solution: regen zproject


include/zproc.h
```diff
@@ -124,11 +124,6 @@ CZMQ_EXPORT int
 CZMQ_EXPORT void *
     zproc_actor (zproc_t *self);
 
-//  *** Draft method, for development use, may change without warning ***
```
bluca: This one should stay - or rather be properly generated

jimklimov:
Nothing pops out at me, as far as recipe bits go. I'm not much into bindings and respective languages though ;)

include/zproc.h
wesyoung: dumb question- shouldn't something like this be in ~platform.h ?

bluca: yep - not sure why it's there

wesyoung: @vyskocilm would defining this here: https://github.com/zeromq/zproject/blob/a7c75864064502de5fbc9fe2c996d03f7e4f0e20/zproject_vs20xx.gsl#L169

instead of include/zproc.h make more sense? that way it's also defined across all zmq projects from here too. can you test that to make sure it fits the problem you were solving?

(i haven't done this on windows- so bare with me, i don't wanna stomp on your fix by a zproject regen.. and i need to PR some changes to the zyre API which requires these to be somewhat in sync)

cmake-build-debug/czmq.cbp
```diff
@@ -0,0 +1,875 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
wesyoung: aye- i'm combing through these.. CLion does some odd stuff when you don't kill cmake (in it) and zproject had some new cmake stuff in it.. learning the differences :)

PR #1853, opened 2018-02-12
Windows: Use SIGTERM instead of SIGKILL

diorcety:


sappo:
Thanks for your PR. For future contributions please follow our contributions guideline and describe the `problem` your commit solves and the `solution` you did implement.

So what's the problem with SIGKILL on windows and why use SIGTERM instead?

bluca:
> So what's the problem with SIGKILL on windows and why use SIGTERM instead?

SIGKILL doesn't exist on Windows, so it had to be redefined, but it was in a generated header which kept being stomped on by zproject, breaking the build

PR #1854, opened 2018-02-12
Problem: zproc_run returns before execve is called

vyskocilm:
Solution: create new execpair zeromq pair, which is used to make
zproc_run synchronized with internal actor. This is the best
approximation of making sure zproc_run ends just before
execve/CreateProcess is called.

### DIFF
--- a/include/zproc.h
+++ b/include/zproc.h
@@ -95,7 +95,7 @@ CZMQ_EXPORT void *
     zproc_stderr (zproc_t *self);
 
 //  *** Draft method, for development use, may change without warning ***
-//  Starts the process.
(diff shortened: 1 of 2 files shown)


include/zproc.h
```diff
@@ -95,7 +95,7 @@ CZMQ_EXPORT void *
     zproc_stderr (zproc_t *self);
 
 //  *** Draft method, for development use, may change without warning ***
-//  Starts the process.
+//  Starts the process, return just before execve/CreateProcess
```
bluca: note that if you don't add this to the api/zproc.api file it will be overwritten

PR #1855, opened 2018-02-12
Problem: zproc reexports a lot of symbols from zsys only

vyskocilm:
Solution: ... the original problem to solve was
7d051910f: no access to zsys methods in generated bindings

However in the meanwhile all reexported zproc methods are described in
zsys API model and available for language bindings. As zproc is DRAFT
class, unlike stable zsys, I'd say such reexport has no meaning and can
be dropped from czmq.

+ regenerate with recent zproject


vyskocilm:
Note: I intentionally kept zproc_interrupted function, as there is no equivalent in zsys atm. However I believe that `zsys_is_interrupted` should be added to `zsys` and zproc will pivot as sub-process handling class.

bluca:
Note that you need to check in the new include/Makefile.am

bluca:
> However I believe that zsys_is_interrupted should be added to zsys and zproc will pivot as sub-process handling class.

Makes sense, but it's not a function, can we do static volatile variables from zproject?

vyskocilm:
@bluca oh - I overlooked new file, thanks 👍

About `zsys_interrupted` - I don't think it's good idea to export `volatile` variables to bindings. I know that some of ffi interfaces (Python cffi) supports that, but some does not (lua jit FFI). I'd rather add two methods to get/set a status of `zsys_interrupted` and let C compiler to deal with `volatile` semantics.

bluca:
A few wrappers were removed from the body but from the API, fixed in: https://github.com/zeromq/czmq/pull/1856

PR #1856, opened 2018-02-12
Problem: JNI and Python CI runs broken

bluca:
Solution: fix them


vyskocilm:
Thanks. I was sure I dropped them from API model as well.

PR #1858, opened 2018-02-13
Don't define strtoull and atoll on MINGW environment

diorcety:
Same issue here: https://github.com/flavio/qjson/issues/26

### DIFF
--- a/include/czmq_prelude.h
+++ b/include/czmq_prelude.h
@@ -535,8 +535,10 @@ typedef struct {
 #   if (!defined (__cplusplus) && (!defined (inline)))
 #       define inline __inline
 #   endif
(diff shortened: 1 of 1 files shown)


vyskocilm:
Hi - please follow the Contributing guidelines and write commit messages in Problem/Solution format. Sometimes it's not obvious for us to know which problem pull request solve and why the solution look like that.

> Problem: can't build czmq with MINGW
> Solution do not define strtoull and atoll on MINGW

diorcety:
@vyskocilm This is the case in the commit message :)

sappo:
@diorcety we put the problem statement on the first line of the commit message. I guess the contribution guide could be more clear about this.

PR #1859, opened 2018-02-13
Problem: zsys_interrupted not available from lang bindings

vyskocilm:
Solution: draft two new `zsys` methods, clean `zproc` class completely and regenerate language bindings to get new methods there.

Follow up of https://github.com/zeromq/czmq/pull/1855

cc @bluca


bluca:
Mmmh need to brush up that C standard and check - will the fact that the variable is accessed through a shared object's function call change the semantics w.r.t. cache coherency and so on?

PR #1862, opened 2018-02-15
fix windows build: explicitly use non unicode versions for CreateProc…

hambre:
…ess and STARTUPINFO

# Pull Request Notice

Compiling with Visual Studio 2015 fails in function s_zproc_execve in file zproc.c.
As the _UNICODE preprocessor macro is defined, CreateProcess and STARTUPINFO automatically resolve
to their unicode versions CreateProcessW and STARTUPINFOW. However CreateProcessW does not work 
with the "commandline" parameter of type char*.
Explicitly using the non-unicode version of CreateProcess and STARTUPINFO fixes the comilation error.

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -513,7 +513,7 @@ s_zproc_execve (zproc_t *self)
 {
     assert (self);
 #if defined(__WINDOWS__)
-    STARTUPINFO siStartInfo;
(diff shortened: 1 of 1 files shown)


vyskocilm:
cc @karolhrdina @thalman - I hope this is correct way to go.

PR #1863, opened 2018-02-15
problem: travis doesnt check zproject

wesyoung:
solution: merge zproject checks from zyre/.travis.yml into czmq/.travis.yml

ironically the zproject check currently fails because of what looks like a manual work-around in the java bindings? (maybe that's a good thing to remind us it's there?)

https://travis-ci.org/wesyoung/czmq/jobs/341938641

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,9 +44,13 @@ env:
     - BUILD_TYPE=stable_zmq ZMQ_REPO=zeromq4-x
     - BUILD_TYPE=stable_zmq ZMQ_REPO=zeromq4-1
     - BUILD_TYPE=cmake
+    - BUILD_TYPE=check_zproject
     - BUILD_TYPE=bindings BINDING=python
 
 matrix:
+  exclude:
+  - os: osx
+    env: BUILD_TYPE=check_zproject
   include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -79,6 +83,7 @@ matrix:
         - sourceline: 'deb http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_14.04/ ./'
           key_url: 'http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_14.04/Release.key'
         packages:
+        - zproject
```
bluca: This installs it only in the ASAN job, not sure it's what you intended?

I'd suggest to add a new -env stanza instead

wesyoung: i... dunno? just trying to mirror what zyre/.travis had to make sure i didn't break something on someone. pulled it, re-running travis job see how it plays out.

bluca:
> ironically the zproject check currently fails because of what looks like a manual work-around in the java bindings? (maybe that's a good thing to remind us it's there?)

Yeah that's a known issue, no solution yet unfortunately

PR #1864, opened 2018-02-16
Remove CZMQ_EXPORT for function definitions

hambre:
When trying to build on windows with the draft-api disabled, compilation fails for multiple reasons.

- some functions definitions in zsys.c have the CZMQ_EXPORT macro which conflicts in non-draft builds with the CZMQ_PRIVATE declaration of these functions in czmq_classes.h
- declaration of structs ztrie_t, zproc_t, zargs_t and ztimerset_t are disables in czmq_library.h for non-draft builds causing comilation errors in ztrie.c, zproc.c, zargs.c and ztimerset.c

For the second point I disabled the respective classes completely using the CZMQ_BUILD_DRAFT_API macro as they were not declared as private classes in non-draft builds.

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -311,7 +311,7 @@ zproc_set_args (zproc_t *self, zlist_t **args_p) {
 
 //  Setup the command line arguments, the first item must be an (absolute) filename
 //  to run. Variadic function, must be NULL terminated.
-CZMQ_EXPORT void
+void
     zproc_set_argsx (zproc_t *self, const char *args, ...)
 {
     assert (self);
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -1513,7 +1513,7 @@ zsys_max_msgsz (void)
 //  Check if default interrupt handler of Ctrl-C or SIGTERM was called.
 //  Does not work if ZSYS_SIGHANDLER is false and code does not call
 //  set interrupted on signal.
-CZMQ_EXPORT bool
+bool
     zsys_is_interrupted (void)
 {
     return zsys_interrupted != 0;
@@ -1524,7 +1524,7 @@ CZMQ_EXPORT bool
 //  Set interrupted flag. This is done by default signal handler, however
 //  this can be handy for language bindings or cases without default
 //  signal handler.
-CZMQ_EXPORT void
+void
     zsys_set_interrupted (void)
 {
     zctx_interrupted = 1;


bluca:
Thanks for the PR, but sorry, but this won't work - as you see in the CI it's all failing.

Shouldn't the proper fix to make the CZMQ_EXPORT/PRIVATE macros work properly on Windows?

hambre:
I do not think it has something to do with the export macros.
The CI is failing because gcc in pedantic mode complains about empty translation units due to my disabling of the complete classes. I could move the #ifdef CZMQ_BUILD_DRAFT_API to just below the struct definitions or add a dummy symbol so that the translation unit is not empty.

bluca:
> I do not think it has something to do with the export macros.

So what is failing exactly on Windows? And what's the configuration? As you can see at this link Appveyor is building fine on Windows at the moment: https://ci.appveyor.com/project/zeromq/czmq/build/build-1270

> The CI is failing because gcc in pedantic mode complains about empty translation units due to my disabling of the complete classes. I could move the #ifdef CZMQ_BUILD_DRAFT_API to just below the struct definitions or add a dummy symbol so that the translation unit is not empty.

I'm afraid these symbols must still be available for internal use - it's intended that they are not completely compiled out - that's why there are the PRIVATE/EXPORT flags. So that will just fail in different ways.

bluca:
That said though, you are right that the CZMQ_EXPORT attributes should be removed from the definitions - they should only be in the headers.

So if you rebase to remove the ifdefs I'll happily merge, and thanks for spotting these errors.

hambre:
-------- Weitergeleitete Nachricht --------
Betreff: 	Re: [zeromq/czmq] Fix window compilation for non-draft builds 
(#1864)
Datum: 	Fri, 16 Feb 2018 12:27:13 +0100
Von: 	Jan Hambrecht <jan.hambrecht@gmx.net>
An: 	Luca Boccassi <notifications@github.com>



Am 16.02.2018 um 12:18 schrieb Luca Boccassi:
>
>     I do not think it has something to do with the export macros.
>
> So what is failing exactly on Windows? And what's the configuration? 
> As you can see at this link Appveyor is building fine on Windows at 
> the moment: https://ci.appveyor.com/project/zeromq/czmq/build/build-1270
>
Yes it compiles just fine when building with the draft api enabled. 
These fixes attempt to fix building with the draft api disabled.
Disabling the draft api causes the declaration of structs ztrie_t, 
zproc_t, zargs_t and ztimerset_t to be disabled in czmq_library.h 
causing comilation errors in ztrie.c, zproc.c, zargs.c and ztimerset.c 
as these structs are only defined and not declared there.
Is building with the draft api disabled is properly working on linux?
>
>     The CI is failing because gcc in pedantic mode complains about
>     empty translation units due to my disabling of the complete
>     classes. I could move the #ifdef CZMQ_BUILD_DRAFT_API to just
>     below the struct definitions or add a dummy symbol so that the
>     translation unit is not empty.
>
> I'm afraid these symbols must still be available for internal use - 
> it's intended that they are not completely compiled out - that's why 
> there are the PRIVATE/EXPORT flags. So that will just fail in 
> different ways.
>
Are you sure? I searched for these symbols and they seem to be only used 
in the bindings which does not count as internal use?

hambre:
Another option would be to just add the declaration of these structs to czmq_classes.h if the draft api is disabled. This would be less invasive and would probably work too. Opinions?

bluca:
It is working fine on Linux, OSX and all other unixes were we test - that's why I was suggesting the problem is most likely in the Windows implementation of those macros.

In fact even the EXPORT in the declaration does not have any effect on *nix:

```
$ objdump -T src/.libs/libczmq.so.4.1.1 | grep zsys_is_inter
$
```

But we should take it out anyway - so if you could please do the rebase, we can merge it and tackle the problem one small issue at a time.

hambre:
Ok, done.

I also investigated some more the other build failures for non-draft builds on windows. It seems that when compiling on linux with autotools or cmake (and drafts disabled), the resulting makefile will not have zargs.c, zproc.c, ztimerset.c and ztrie.c as sources files. So that is why compilation on linux works. The visual studio files does contain these files as source in any case (there does not seem to be support to contionally include source files) so that is why building with drafts disabled fails.

Issue #1807, opened 2017-12-01
CZMQ won't build with DRAFT API


paddor:
I just pulled ZMQ HEAD and CZMQ HEAD into /usr/local/src. I compiled and installed ZMQ with:

```
./autogen.sh
./configure --enable-drafts=yes --with-libsodium --without-docs
make
make check
sudo make install
```

Then I did almost the same for CZMQ:
```
./autogen.sh
./configure --enable-drafts=yes  --without-docs
make
make check
sudo make install
```

Here's the output of CZMQ's `configure` script:

```
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether UID '1000' is supported by ustar format... yes
checking whether GID '1000' is supported by ustar format... yes
checking how to create a ustar tar archive... gnutar
checking whether make supports nested variables... (cached) yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking for gcc option to accept ISO C99... none needed
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking how to print strings... printf
checking for a sed that does not truncate output... /bin/sed
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for fgrep... /bin/grep -F
checking for ld used by gcc... /home/p/.linuxbrew/bin/ld
checking if the linker (/home/p/.linuxbrew/bin/ld) is GNU ld... yes
checking for BSD- or MS-compatible name lister (nm)... /home/p/.linuxbrew/bin/nm -B
checking the name lister (/home/p/.linuxbrew/bin/nm -B) interface... BSD nm
checking whether ln -s works... yes
checking the maximum length of command line arguments... 1572864
checking how to convert x86_64-unknown-linux-gnu file names to x86_64-unknown-linux-gnu format... func_convert_file_noop
[...]

PR #1865, opened 2018-02-19
Problem: wrong errno set when feature is missing

paddor:
Solution: set errno to ENOTSUP

### DIFF
--- a/src/zsock.c
+++ b/src/zsock.c
@@ -2174,6 +2174,16 @@ zsock_test (bool verbose)
     zsock_destroy (&client);
     zsock_destroy (&server);
 
+#else
+    errno = 0;
+    zsock_t* server = zsock_new_server (NULL);
+    assert(server == NULL);
+    assert(errno == ENOTSUP);
+
+    errno = 0;
+    zsock_t* client = zsock_new_client (NULL);
+    assert(client == NULL);
+    assert(errno == ENOTSUP);
 #endif
 
 #ifdef ZMQ_SCATTER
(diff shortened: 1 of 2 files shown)


bluca:
Thanks for the pr but unfortunately these are API breakages so they can only be done for draft ones - could you please rebase and take out the changes to the stable APIs? Thanks

paddor:
DISH/SERVER/... are stable now? Or which ones do you mean?

As for the cases where `NULL` is returned, any `errno` value read by client code was a previously set one.

paddor:
I briefly stumbled across this problem in #1807 (see bottom of issue description). And today I wasted time because my tests got an EINVAL when setting the group on a frame until I realized that CMZQ got installed without DRAFT APIs again (this time with brew on Mac, which is another problem). 

Sorry but EINVAL just doesn’t make sense in these cases and not setting an errno at all (in the constructors) is no help when debugging.

bluca:
> I just checked the API files. Seems like SERVER/CLIENT/RADIO/DISH are still draft.

Yes those changes are all fine - but xpub/xsub/stream are stable so we cannot change the errno. I know enotsup makes way more sense but it's too late I'm afraid.

src/zsock.c
```diff
@@ -2165,6 +2174,16 @@ zsock_test (bool verbose)
     zsock_destroy (&client);
     zsock_destroy (&server);
 
+#else
+    errno = 0;
+    zsock_t* server = zsock_new_server (NULL);
```
bluca:
```
src/zsock.c: In function ‘zsock_test’:
src/zsock.c:2179:14: error: redefinition of ‘server’
     zsock_t* server = zsock_new_server (NULL);
              ^
src/zsock.c:1924:14: note: previous definition of ‘server’ was here
     zsock_t *server = zsock_new (ZMQ_DEALER);
```

This breaks the tests, could you please fix it? Thanks!

paddor:
Thanks for the clarification. Updated.

I'm curious though: How is an undefined value in `errno` part of the stable API? To my understanding, it used to be whatever it was set to in a previous call. Are you thinking of programs depending on that undefined value, as long as it is deterministic?

src/zsock.c
```diff
@@ -321,6 +322,7 @@ zsock_new_xsub_checked (const char *endpoints, const char *filename, size_t line
         zsock_destroy (&sock);
     return sock;
 #else
+    errno = ENOTSUP;
```
bluca: This one too please :-)

src/zsock.c
```diff
@@ -363,6 +365,7 @@ zsock_new_stream_checked (const char *endpoints, const char *filename, size_t li
         zsock_destroy (&sock);
     return sock;
 #else
+    // errno = ENOTSUP;      // too late, stable API would be broken
```
bluca: Ditto

src/zsock.c
paddor: Sorry, missed this one. Updated.

bluca:
> I'm curious though: How is an undefined value in errno part of the stable API? To my understanding, it used to be whatever it was set to in a previous call. Are you thinking of programs depending on that undefined value, as long as it is deterministic?

Going from "does not set an error" to "sets this error in this case" is an interface change that would cause users of that api to have to at the very least re evaluate their usage when they read the changelog, which is something that a stable library should not cause

src/zsock.c
```diff
@@ -1912,17 +1921,17 @@ zsock_test (bool verbose)
     assert (port >= 60000 && port <= 60500);
 
     //  Test zsock_attach method
-    zsock_t *server = zsock_new (ZMQ_DEALER);
```
bluca: Sorry to be a pain in the back but CI compilation is still broken:

```
src/zsock.c: In function 'zsock_test':
src/zsock.c:2111:5: error: 'server' undeclared (first use in this function)
     server = zsock_new (ZMQ_SERVER);
```

There's more usage of that variable below, so perhaps it's easier to use different names in the new snippet? Up to you though

paddor: No problem. Fixed and running `make && make check` this time. Stay tuned...

paddor: Okay, this time it should be good. Sorry about this, I'm too used to Ruby. 😆

paddor:
Thanks!

bluca:
Thanks to you for the fixes and being patient :)

PR #1868, opened 2018-02-22
PROBLEM: Need getter for zproc cmd line arguments

karolhrdina:
SOLUTION: Implement it.

### DIFF
--- a/src/czmq_selftest.c
+++ b/src/czmq_selftest.c
@@ -21,55 +21,64 @@
 
 #include "czmq_classes.h"
 
+#include <stdio.h>
+#include <string.h>
+
 typedef struct {
-    const char *testname;
-    void (*test) (bool);
+    const char *testname;           // test name, can be called from command line this way
+    void (*test) (bool);            // function to run the test (or NULL for private tests)
+    bool stable;                    // true if class is declared as stable
+    bool pub;                       // true if class is declared as public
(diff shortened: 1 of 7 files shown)


bluca:
looks like the *nix builds are broken, could you please have a look?

src/czmq_selftest.c
```diff
@@ -97,21 +106,52 @@ test_runall (bool verbose)
...
     printf ("Running czmq selftests...\n");
-    for (item = all_tests; item->test; item++)
-        item->test (verbose);
+    for (item = all_tests; item->testname; item++) {
+        if (strcmp (item->testname, "private_classes") == 0)
+            continue;
+        if (!item->subtest)
+            item->test (verbose);
+        else
+            czmq_private_selftest (verbose, item->subtest);
```
bluca: I think this is the problem - needs to be ifdefd

src/czmq_selftest.c
```diff
@@ -205,7 +210,10 @@ main (int argc, char **argv)
 
     if (test) {
         printf ("Running czmq test '%s'...\n", test->testname);
-        test->test (verbose);
+        if (!test->subtest)
+            test->test (verbose);
+        else
+            czmq_private_selftest (verbose, test->subtest);
```
bluca: this too

Issue #1867, opened 2018-02-22
Double free when custom zactor destructor uses bsend/brecv


vyskocilm:
How to reproduce
```
diff --git a/src/zactor.c b/src/zactor.c
index cac3c371..0daa5c85 100644
--- a/src/zactor.c
+++ b/src/zactor.c
@@ -300,8 +300,32 @@ KTHXBAI_actor (zsock_t *pipe, void *args) {
 static void
 KTHXBAI_destructor (zactor_t *self) {
     assert (self);
-    if (zstr_send (self->pipe, "$KTHXBAI") == 0)
-        zsock_wait (self->pipe);
+    if (zstr_send (self, "$KTHXBAI") == 0)
+        zsock_wait (self);
+}
+
+static void
+BSEND_actor (zsock_t *pipe, void *args) {
+
+    zsock_signal (pipe, 0);
+    while (!zsys_interrupted) {
+        char *str = NULL;
+        void *ptr = NULL;
+        int r = zsock_brecv (pipe, "sp", &str, &ptr);
+        if (r == -1)
+            break;
+        int done = streq (str, "$TERM");
+        zstr_free (&str);
+        if (done)
+            break;
+    }
+}
+
+static void
+BSEND_destructor (zactor_t *self) {
+    assert (self);
+    if (zsock_bsend (self, "sp", "$TERM", NULL) == 0)
+        zsock_wait (self);
 }
 
 //  --------------------------------------------------------------------------
@@ -329,6 +353,13 @@ zactor_test (bool verbose)
     zactor_set_destructor (KTHXBAI, KTHXBAI_destructor);
     zactor_destroy (&KTHXBAI);
 
+    // custom destructor
+    // KTHXBAI_actor ends on "$KTHXBAI" string
+    zactor_t *BSEND = zactor_new (BSEND_actor, NULL);
+    assert (BSEND);
+    // which is the one sent by KTHXBAI_destructor
+    zactor_set_destructor (BSEND, BSEND_destructor);
+    zactor_destroy (&BSEND);
 #if defined (__WINDOWS__)
     zsys_shutdown();
 #endif
```

`zactor` selftest then ends on SIGABRT

```
Core was generated by `/home/mvyskocil/projects/zeromq/czmq/src/.libs/czmq_selftest'.                                                                                        
Program terminated with signal SIGABRT, Aborted.                                                                                                                             
[...]

PR #1870, opened 2018-02-23
Problem: custom destructor crashes unit test

vyskocilm:
Solution: see commits

fixes https://github.com/zeromq/czmq/issues/1867


bindings/jni/src/main/java/org/zeromq/czmq/Zconfig.java
```diff
@@ -229,10 +229,9 @@ public void removeSubtree () {
     }
     /*
     Destroy node and subtree (all children)
-    WARNING: manually fixed void -> long
```
bluca: please don't overwrite this - it will break the JNI bindings

vyskocilm: reverted the change

PR #1871, opened 2018-02-26
clang-format custom target failes to build if building with libzmq

TomzOk:
Problem:
When building both libzmq and czmq with cmake, cmake fails because of the clang-format target being defined twice.  This PR renames the clang-format target to clang-format-czmq.  You can then make target with make clang-format-czmq.

(Also note that running clang-format target seems to break the code - it does not seem like this target is maintained or used for code checkins so maybe should remove)

### DIFF
--- a/builds/cmake/Modules/ClangFormat.cmake
+++ b/builds/cmake/Modules/ClangFormat.cmake
@@ -19,7 +19,7 @@ if("${CLANG_FORMAT}" STREQUAL "")
 endif()
 
 add_custom_target(
-        clang-format
+        clang-format-czmq
         COMMAND ${CLANG_FORMAT} -style=file -i ${ALL_SOURCE_FILES}
 )
 
@@ -31,14 +31,14 @@ endfunction()
 configure_file(builds/cmake/clang-format-check.sh.in clang-format-check.sh @ONLY)
 
 add_custom_target(
-        clang-format-check
+        clang-format-czmq-check
         COMMAND chmod +x clang-format-check.sh
         COMMAND ./clang-format-check.sh
         COMMENT "Checking correct formatting according to .clang-format file using ${CLANG_FORMAT}"
 )
 
 add_custom_target(
-        clang-format-diff
+        clang-format-czmq-diff
         COMMAND ${CLANG_FORMAT} -style=file -i ${ALL_SOURCE_FILES}
         COMMAND git diff ${ALL_SOURCE_FILES}
         COMMENT "Formatting with clang-format (using ${CLANG_FORMAT}) and showing differences with latest commit"


bluca:
This needs to be done in https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl or it will be overwritten - @jimklimov FYI

bluca:
Although I'm not sure why it would be a problem - I didn't think these files would be exported, I thought they would stay private

TomzOk:
If the change will be overwritten by the project tool then why did you merge it?

That is confusing

bluca:
Because it won't happen for a while as it's done manually, and in the meanwhile it can fix the problem

jimklimov:
I'd say, if clang-format does break the code, it should be brought up wuth the clang project. And attn @sigiesec as he proposed this tool as having the benefit over common scripted style tooks - this one is based on an actual compiler.

GSL renaming might be done with adding the $(project) name to the target, but it sort of defeats the purpose (having all projects behave the same).

I am uncertain about the original premise - "building libzmq and czmq"... all other same-named targets would collide? Or is it "with cmake" which I don't use, and don't even know for sure if it defines its own rules or uses same Makefile.am et al as autotools? If the latter, then indeed we end up having it defined twice, so the cmake variant can go away (we then cover both toolkits with `*.am` files) :)

TomzOk:
Regarding exports,

I added both czmq and zmq projects with cmakes  add_subdirectory() method so that's why they collide. 
 
The clang format target is the only collision, so it's convenient to fix. However if you forsee other collisions than it's better to not fix and require cmake dependants import both projects with cmakes add_externalProject () for each.

Also since clang format target seems broken it's also better to just remove.

jimklimov:
Can you test if you have the same collision building the target in just one project using cmake? Then it is indeed likely due to us having added the same recipes in both cmake and autotools GSL scripts.

Otherwise, if the collision is due to two projects driven by your routine - literally everything would collide (all, memcheck, etc.)

bluca:
@jimklimov - autotools and cmake are completely different and separete

@TomzOk - including everything in the same cmake project is not really supported - why is that necessary? We have pkgconfig and all the cmake-specific findpackage stuff as well

TomzOk:
Cmake uses it's own makefiles and directories so there wouldn't be colliding with auto tools. Each project build fine for me with cmake and auto tools when One at time or when importing with cmake external_project_add().  You just cant use add_subdirectory() to pull in both projects. When you do use add_subdirectory to import both projects, only clang format target collides. With the single name space fix you can now also import both projects with add_subdirectory ().

You can also not fix and just maybe note how to import both projects with external_project_add().

TomzOk:
Pulling into same project was convenient is all, not really necessary. And it works with little tweak.

I guess if you are developer you can run tests and builds from one project instead of two. Other than that it sounds like not maintainable.

I understand the maintenance of that might not be worth it though.

TomzOk:
Just as side note - when running the clang-format target it seems to break the code.  And a git diff after the formatting shows about 30K line changes so the clang-format target does not look used...  This was noted above but just for completeness here was the compiler error.

``In file included from .../czmq/src/zgossip_msg.c:32:0:
.../czmq/src/./zgossip_msg.h:36:1: error: unknown type name ‘CZMQ_PRIVATE’
 CZMQ_PRIVATE zgossip_msg_t* zgossip_msg_new(void)'
``

PR #1874, opened 2018-03-05
Support zmq zero copy option

skaes:
This PR adds support for the new ZMQ_ZERO_COPY_RECV option of libzmq.

It is needed because zmq contexts are handled internally by czmq.

### DIFF
--- a/api/zsys.api
+++ b/api/zsys.api
@@ -316,6 +316,18 @@
         <return type = "integer" />
     </method>
 
+    <method name = "set zero copy recv" singleton = "1" state = "draft" >
+        Configure whether to use zero copy strategy in libzmq. If the environment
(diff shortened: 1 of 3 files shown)


bluca:
We use a code generation tool in this project for bindings and so on. You don't have to use it, but please add the definition of the new APIs in the api/zsys.api file. You can copy&paste this one for example:

https://github.com/zeromq/czmq/blob/master/api/zsys.api#L308

bluca:
also use ```state = "draft"``` as it's a new API

bluca:
Great, thank you - will take care of regenerating the bindings

skaes:
@bluca Thx a lot. If you could show me how to generate the bindings, I can do it on my own the next time.

PR #1876, opened 2018-03-07
Problems: atexit called after zmq static object destruction causes crash, out of date with zproject

bluca:
Solutions: see commits

See: https://github.com/zeromq/libzmq/pull/2971


jimklimov:
Existing travis red marks confirmed to be non-issues (a known manual workaround, and a zmq2 test failing since day X).

PR #1889, opened 2018-04-06
Problem: zproc stdin handling is wrong

vyskocilm:
Solution: see commits and added test case


vyskocilm:
There is some memory error reported by AddressSanitizer - I'll try to figure it out.

PR #1893, opened 2018-04-12
Problem: zdir_new() is not careful with its inputs

jimklimov:
Solutions:
* check for NULL arguments before using them
* check for empty `path` when it is the full path (parent is `-`)
** note: empty path and empty parent implicitly means rootfs here; should it?.. actually empty parent seems not checked at all
** note: empty path and non-empty parent resolves into parent, this looks right
* do not cut the trailing slash of the root dir (`path == /`)
* cut multiple trailing slashes, not just one


jimklimov:
Yes, splitting the PR. Seems there are human decisions to be made about validity of empty/NULL inputs.

vyskocilm:
What problem are you trying to solve by cutting last trailing slash? I fail to see how `opendir ("path/////");` can fail. What is worse is it slightly changes behavior of a class declared as stable ...

jimklimov:
Try opening a root path "/" with existing code ;)
The original code just looks if the last char is a slash and zeroes it out. Does two strlen's, does not check if the path is only the slash (or is empty, so using negative array offset). Just as well, it did not chomp multiple trailing slashes, so whatever purpose one cut-off was done for (perhaps, to build a clean pathname that can be suffixed with something else), that goal was not fully achieved with other bad inputs.

PR #1898, opened 2018-04-18
Problem: zproc test commented out

vyskocilm:
Solution: uncomment them

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -1038,7 +1038,7 @@ zproc_test (bool verbose)
     
     int r = zproc_run (self);
     assert (r == 0);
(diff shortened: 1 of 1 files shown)


bluca:
The address sanitizer reports a buffet overflow in zproc could you please have a look?

vyskocilm:
@bluca finally found why it dissapeared before as I've commented out the code :-D

The bug was actually in the test code as I was constructing frame with two bytes more than was actual length of variable. Excuse the old C hacker to take shortcuts :)

Should be fixed now.

vyskocilm:
There is `4-1` fail in `zgossip` I believe this is not related. The rest looks good

PR #1900, opened 2018-04-18
Problem: UDP socket descriptor leak with exec'd processes

### LINKED ISSUE #1899: Problem: UDP socket descriptor leak with exec'd processes
The `zsys_udp_new()` method does not set the `O_CLOEXEC` flag on the socket descriptor, which leads to the descriptor leaking if the process subsequently performs an `exec()`. This is a particular issue for the `zbeacon` class as the flag cannot be set by client code either, since the UDP socket descriptors are private to the class.

stephenprocter:
Solution: Set O_CLOEXEC on UDP socket descriptors. Fixes #1899

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -980,14 +980,35 @@ zsys_udp_new (bool routable)
     //  We haven't implemented multicast yet
     assert (!routable);
     SOCKET udpsock;
+    int type = SOCK_DGRAM;
+#if defined ZMQ_HAVE_SOCK_CLOEXEC
+    //  Ensure socket is closed by exec() functions.
+    type |= SOCK_CLOEXEC;
+#endif
+
     if (zsys_ipv6 ())
-        udpsock = socket (AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
+        udpsock = socket (AF_INET6, type, IPPROTO_UDP);
     else
-        udpsock = socket (AF_INET, SOCK_DGRAM, IPPROTO_UDP);
+        udpsock = socket (AF_INET, type, IPPROTO_UDP);
     if (udpsock == INVALID_SOCKET) {
         zsys_socket_error ("socket");
         return INVALID_SOCKET;
     }
+
+    //  If there's no SOCK_CLOEXEC, let's try the second best option. Note that
+    //  race condition can cause socket not to be closed (if fork happens
+    //  between socket creation and this point).
+#if !defined ZMQ_HAVE_SOCK_CLOEXEC && defined FD_CLOEXEC
+    if (fcntl (udpsock, F_SETFD, FD_CLOEXEC) == SOCKET_ERROR)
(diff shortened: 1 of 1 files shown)


vyskocilm:
While I am not sure that the "second best option" is useful at all as having platform without SOCK_CLOEXEC and with FD_CLOEXEC is very improbable, I think we can merge this one.

```
+    //  If there's no SOCK_CLOEXEC, let's try the second best option. Note that
+    //  race condition can cause socket not to be closed (if fork happens
+    //  between socket creation and this point).
```

src/zsys.c
```diff
@@ -980,14 +980,35 @@ zsys_udp_new (bool routable)
     //  We haven't implemented multicast yet
     assert (!routable);
     SOCKET udpsock;
+    int type = SOCK_DGRAM;
+#if defined ZMQ_HAVE_SOCK_CLOEXEC
```
bluca: this is not defined anywhere - the logic to detect that needs to be added to zproject

src/zsys.c
```diff
@@ -980,14 +980,35 @@ zsys_udp_new (bool routable)
...
+    //  If there's no SOCK_CLOEXEC, let's try the second best option. Note that
+    //  race condition can cause socket not to be closed (if fork happens
+    //  between socket creation and this point).
+#if !defined ZMQ_HAVE_SOCK_CLOEXEC && defined FD_CLOEXEC
+    if (fcntl (udpsock, F_SETFD, FD_CLOEXEC) == SOCKET_ERROR)
+        zsys_socket_error ("fcntl (FD_CLOEXEC)");
+#endif
+
+    //  On Windows, preventing sockets to be inherited by child processes.
+#if defined ZMQ_HAVE_WINDOWS && defined HANDLE_FLAG_INHERIT
```
bluca: this is not defined either - it's CZMQ_

src/zsys.c
vyskocilm: I don't think that it's worth to add autoconf machinery for `SOCK_CLOEXEC`, it's just a macro. I fixed the things in PR#1901

vyskocilm: BTW: thanks for a review!

bluca: It is necessary - the definition might be available, but it might not work. That's the case on GNU/Hurd for example. See libzmq/acinclude.m4 and libzmq/builds/cmake/Modules/ZMQSourceRunChecks.cmake

src/zsys.c
```diff
@@ -980,14 +980,35 @@ zsys_udp_new (bool routable)
...
+    BOOL brc = SetHandleInformation ((HANDLE) udpsock, HANDLE_FLAG_INHERIT, 0);
+    win_assert (brc);
```
bluca: @stephenprocter this breaks the windows build:

```
C:\projects\czmq\src\zsys.c(1001): error C3861: 'win_assert': identifier not found [C:\projects\build_czmq\czmq.vcxproj]
```

PR #1901, opened 2018-04-18
Problem: zsys_udp_new reference internal ZMQ macros

vyskocilm:
Solution: use external ones
fixes: https://github.com/zeromq/czmq/issues/1899 https://github.com/zeromq/czmq/pull/1900

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -981,7 +981,7 @@ zsys_udp_new (bool routable)
     assert (!routable);
     SOCKET udpsock;
     int type = SOCK_DGRAM;
-#if defined ZMQ_HAVE_SOCK_CLOEXEC
+#if defined (SOCK_CLOEXEC)
     //  Ensure socket is closed by exec() functions.
     type |= SOCK_CLOEXEC;
 #endif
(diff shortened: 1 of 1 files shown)


src/zsys.c
```diff
@@ -995,16 +995,8 @@ zsys_udp_new (bool routable)
...
-    //  race condition can cause socket not to be closed (if fork happens
-    //  between socket creation and this point).
-#if !defined ZMQ_HAVE_SOCK_CLOEXEC && defined FD_CLOEXEC
-    if (fcntl (udpsock, F_SETFD, FD_CLOEXEC) == SOCKET_ERROR)
-        zsys_socket_error ("fcntl (FD_CLOEXEC)");
-#endif
-
     //  On Windows, preventing sockets to be inherited by child processes.
-#if defined ZMQ_HAVE_WINDOWS && defined HANDLE_FLAG_INHERIT
+#if defined (__WINDOWS__) && defined (HANDLE_FLAG_INHERIT)
```
bluca: There's CZMQ_HAVE_WINDOWS please use that one

vyskocilm: Fixed in latest commit

src/zsys.c
```diff
@@ -981,7 +981,7 @@ zsys_udp_new (bool routable)
     assert (!routable);
     SOCKET udpsock;
     int type = SOCK_DGRAM;
-#if defined ZMQ_HAVE_SOCK_CLOEXEC
+#if defined (SOCK_CLOEXEC)
```
bluca: @vyskocilm this will breaks some platforms - it needs to be done in cmake and autoconf like in libzmq

PR #1909, opened 2018-05-08
Problem: We can't wait for process with timeout

thalman:
Solution: Rework zproc_wait to use timeout, use alarm () to exit from waitpid call
Create zproc_shutdown function that does the typical usecase TERM, wait, KILL

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -287,7 +287,7 @@ zproc_destroy (zproc_t **self_p) {
     assert (self_p);
     if (*self_p) {
         zproc_t *self = *self_p;
-        zproc_wait (self, true);
+        zproc_wait (self, -1);
         zactor_destroy (&self->actor);
 
         if (self->stdinpipe [0] != -1)  close (self->stdinpipe [0]);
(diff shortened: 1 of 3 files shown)


src/zproc.c
```diff
@@ -808,24 +811,55 @@ zproc_wait (zproc_t *self, bool wait) {
...
+        // infinite
+        r = waitpid (self->pid, &status, 0);
+    }
+    else if (timeout == 0) {
+        // no hang
+        r = waitpid (self->pid, &status, WNOHANG);
+    } else {
+        // with timeout
+        // set sigalarm handler
+        struct sigaction myhandler;
```
sappo: Please rename to be more meaningful

sappo:
@thalman the zproc test against latest libzmq are failing. Can you please have a look https://travis-ci.org/zeromq/czmq/jobs/376591885. Thanks!

vyskocilm:
I like new methods, however don't think that using alarm is the sane way of doing this. I still think that zproc should behave more like actor in some way... And timeouts should be implemented by sending messages.

thalman:
Alert approach is definitely not good, I will redo that.

PR #1914, opened 2018-05-10
Problem: Windows build broken by non-existing macro

bluca:
Solution: simply call assert - win_assert is an internal macro of libzmq


bluca:
Fixed by https://github.com/zeromq/czmq/pull/1915

PR #1915, opened 2018-05-10
Problem: win_assert() undefined, breaks Windows build.

stephenprocter:
Solution: replace win_assert() with zsys_socket_error().


bluca:
Thanks!

PR #1918, opened 2018-05-11
[WIP] Problem: it's too easy to write hanging code based on zproc

vyskocilm:
Solution:
 1. never loop in s_fd_in_handler
 2. call new shutdown in destructor

The key is it's much easier to debug prematurerly ended programs than complete hangs ....

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -287,7 +287,7 @@ zproc_destroy (zproc_t **self_p) {
     assert (self_p);
     if (*self_p) {
         zproc_t *self = *self_p;
-        zproc_wait (self, -1);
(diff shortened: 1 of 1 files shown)


bluca:
Seems to still be failing, want to have a look before I merge?

vyskocilm:
Sure, it looks there is some fundamental error. I will be away during weekend, so please be patient with me ☺️

bluca:
Of course, no problem, thanks! I've marked as WIP

vyskocilm:
It looks that @thalman solved the problem in a better way, so closing my PR.

PR #1920, opened 2018-05-11
[WIP] Problem: Waitpid can't be called by user

thalman:
Solution: Move it from zproc_wait to private function,
Check only "running" property in zproc_wait


bluca:
src/selftest-ro/.gitkeep should not be removed, any reason to delete it?

thalman:
@bluca my mistake, it is back now

vyskocilm:
Merged, will fix the invalid read problem ...

PR #1921, opened 2018-05-14
Problem: zproc robustness is low

vyskocilm:
Solution: see individual commits, this is mostly the polishing of changes from https://github.com/zeromq/czmq/pull/1920

Tested locally and worked fine.

### DIFF
--- a/src/zproc.c
+++ b/src/zproc.c
@@ -215,7 +215,6 @@ struct _zproc_t {
     int stdoutpipe [2];     // stdout pipe
     int stderrpipe [2];     // stderr pipe
 
-    zpair_t *execpair;      // pair used to synchronize zproc_run with actor
     zpair_t *stdinpair;     // stdin socketpair
     zpair_t *stdoutpair;    // stdout socketpair
     zpair_t *stderrpair;    // stderr socketpair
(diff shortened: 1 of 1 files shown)


bluca:
There are missing changes to the header from regen:

```
diff --git a/include/zproc.h b/include/zproc.h
index eb18506..b467794 100644
--- a/include/zproc.h
+++ b/include/zproc.h
@@ -122,10 +122,9 @@ CZMQ_EXPORT bool
     zproc_running (zproc_t *self);
 
 //  *** Draft method, for development use, may change without warning ***
-//  wait or poll process status, return return code or ZPROC_RUNNING
-//  The timeout should be zero or greater, or -1 to wait indefinitely.
+//  wait or poll process status, return return code
 CZMQ_EXPORT int
-    zproc_wait (zproc_t *self, int timeout);
+    zproc_wait (zproc_t *self, bool hang);
 
 //  *** Draft method, for development use, may change without warning ***
 //  return internal actor, useful for the polling if process died
@@ -137,12 +136,6 @@ CZMQ_EXPORT void *
 CZMQ_EXPORT void
     zproc_kill (zproc_t *self, int signal);
 
-//  *** Draft method, for development use, may change without warning ***
-//  send SIGTERM signal to the subprocess, wait for grace period and
-//  eventually send SIGKILL
-CZMQ_EXPORT void
-    zproc_shutdown (zproc_t *self, int timeout);
-
 //  *** Draft method, for development use, may change without warning ***
 //  set verbose mode
 CZMQ_EXPORT void
```

vyskocilm:
@bluca did now it - fixed in https://github.com/zeromq/czmq/pull/1922

PR #1923, opened 2018-05-14
Problem: jni broken

vyskocilm:
Solution: return the workaround back

cc @bluca


jimklimov:
Perhaps add a rule in docs that for the time being, `git difftool -y` is a must afyer regen of certain projects, as well as add a comment in file header that this one is not 100% generated, so eyes are really looking early on (for comments near tweaked lines) when they should?

PR #1927, opened 2018-05-23
Problem: zproject-generated files are outdated

gabm:
Solution: update these files to include the latest changes of zproject, mainly the cmake pkg config support


bindings/jni/src/main/java/org/zeromq/czmq/Zconfig.java
```diff
@@ -238,10 +238,9 @@ public void removeSubtree () {
     }
     /*
     Destroy node and subtree (all children)
-    WARNING: manually fixed void -> long
```
bluca: please revert changes to this file

gabm: of course - although its generated by zproject automatically - so maybe we need to fix it there?

bluca: yes - there's an outstanding issue, but nobody has provided a fix until now

bluca:
The cmake build is broken - the new cmake file was not added to EXTRA_DIST in Makefile.am:

```
CMake Error: File /home/travis/build/zeromq/czmq/czmq-4.2.0/builds/cmake/Config.cmake.in does not exist.
CMake Error at /usr/local/cmake-3.9.2/share/cmake-3.9/Modules/CMakePackageConfigHelpers.cmake:312 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  CMakeLists.txt:420 (configure_package_config_file)
```

gabm:
oh I see - I intentionally removed that from my zproject commit oO.. I'll fix it there.. sorry

gabm:
I created a PR here: https://github.com/zeromq/zproject/pull/1089

gabm:
... lets see if it works now..

PR #1934, opened 2018-08-09
Problem: can't easilly create strings with repetitive format arguments

vyskocilm:
Solution: put python-like format functions to czmq

This adds python-like string formatting, where one can use hash and
zconfig to create arbitrary big and complicate strings based on
formatting template.

Reviewed-by: Mike Donovan <mike.donovan@dbvisit.com>


sappo:
Can you please mark the methods as draft in the API.

vyskocilm:
@sappo fixed, it was initially there, but somewhat missed it during all the changes afterwards.

sappo:
thanks

PR #1937, opened 2018-09-25
Update README.md

ngbrown:
Problem: Requiring `apt-get install git-all` is excessive and `libsodium-dev` is also required.

Solution: require only `git` and add `libsodium-dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ To report an issue, use the [CZMQ issue tracker](https://github.com/zeromq/czmq/
 
 To start with, you need at least these packages:
 
-* {{git-all}} -- git is how we share code with other people.
+* {{git}} -- git is how we share code with other people.
 
 * {{build-essential}}, {{libtool}}, {{pkg-config}} - the C compiler and related tools.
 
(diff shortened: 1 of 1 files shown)


sappo:
Hi @ngbrown, sorry for the slow response.
I agree that git-all is excessive but libsodium-dev is optional.

sappo:
I have to correct myself libsodium-dev does nothing for czmq as it is an optional dependency for libzmq

bluca:
> I have to correct myself libsodium-dev does nothing for czmq as it is an optional dependency for libzmq

The example ./configure command for building libzmq below that uses --with-libsodium so I think it's fair to list it - and if not used, it won't hurt, it's available pretty much everywhere anyway :-)

bluca:
@ngbrown thanks for the PR but please note that the README.md file is generated, so you'll want to send another PR to do the same change in README.txt otherwise it will get overwritten

sappo:
> > I have to correct myself libsodium-dev does nothing for czmq as it is an optional dependency for libzmq
> 
> The example ./configure command for building libzmq below that uses --with-libsodium so I think it's fair to list it - and if not used, it won't hurt, it's available pretty much everywhere anyway :-)

Good point

PR #1947, opened 2018-11-11
Problem: no way to send/recv list of strings, as in zproto

somdoron:
Solution: Implement sending and receiving list or strings, as in zproto.


somdoron:
Easier to review the first two commits, as the third one is just the regeneration of bindings.

somdoron:
@bluca done

PR #1954, opened 2018-11-15
Problem: zhttp_client cannot be added to zpoller

somdoron:
Solution: make it an alias to zactor

### DIFF
--- a/src/zhttp_client.c
+++ b/src/zhttp_client.c
@@ -24,12 +24,6 @@
 #include <curl/curl.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


src/zhttp_client.c
```diff
@@ -27,7 +27,6 @@
 //  Structure of our class
 
 struct _zhttp_client_t {
```
bluca: can this be removed now?

somdoron: removed

PR #1959, opened 2018-11-30
use memcpy on unaligned pointer to avoid crash on armv7 (BUS_ADRALN)

hgourvest:
There is a crash in zhash_pack method when writting the 4 bytes corresponding to value length, to unaligned pointer address, this occur on Android armv7. I solve the problem using memcpy instead.

### DIFF
--- a/src/zhash.c
+++ b/src/zhash.c
@@ -701,7 +701,7 @@ zhash_pack (zhash_t *self)
             needle += length;
 
             //  Store value as longstr
-            length = strlen (item->value);
+            length = strlen ((char *) item->value);
             uint32_t serialize = htonl ((uint32_t) length);
             memcpy (needle, &serialize, 4);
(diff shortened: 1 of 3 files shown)


hgourvest:
probably same problem on zlistx_pack and zhashx_pack_own

sappo:
Ah got it after all.

hgourvest:
[zhash] before: 704 after: 706
[zhashx] before: 822 after: 824
[zlistx] before: 674 after: 675

sorry for the spaces, Android studio automatically remove unncecessary space :/

sappo:
Thanks. Don't worry about the spaces my vim configuration does the same
thing ;)

Am Fr., 30. Nov. 2018, 11:40 hat Henri Gourvest <notifications@github.com>
geschrieben:

PR #1965, opened 2018-12-13
Problem: czmq out of sync with zproject

sappo:
Solution: Re-generate!


vyskocilm:
I am afraid the zconfig Java workaround must be added back

bluca:
> I am afraid the zconfig Java workaround must be added back

The JNI test passes - it would be broken if that was still necessary, so I guess it was fixed

PR #1977, opened 2019-01-14
Problem: cannot create zero-copy zframe & zchunk 

somdoron:
Solution: Add constructors that take ownership of the memory

### DIFF
--- a/src/zframe.c
+++ b/src/zframe.c
@@ -40,6 +40,8 @@ struct _zframe_t {
 #ifdef ZMQ_GROUP_MAX_LENGTH
     char group[ZMQ_GROUP_MAX_LENGTH + 1];
 #endif
+    zframe_destructor_fn *destructor;   //  Destructor for memory
+    void * hint;                        //  Hint for destroying the memory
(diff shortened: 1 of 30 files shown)


src/zframe.c
```diff
@@ -40,6 +40,10 @@ struct _zframe_t {
 #ifdef ZMQ_GROUP_MAX_LENGTH
     char group[ZMQ_GROUP_MAX_LENGTH + 1];
 #endif
+#ifdef CZMQ_BUILD_DRAFT_API
```
bluca: Note that it's not necessary to wrap in ifdef the implementation. The symbols will be hidden from the shared library if the API is draft and the build option is not enabled, so it can't be used anyway. This should make the implementation easier and cleaner.

somdoron: Cool, will fix it. I do want to leave it in the other functions as I don't want to change existing stable logic.

somdoron: Fixed, I also fixed the other method eventually, I think it is pretty safe.

PR #1987, opened 2019-02-01
Regenerate CI build recipes with current zproject

jimklimov:
Commented what was customized in the travis script, updated the solution with recent Jenkinsfile patterns improvements, etc.

Note that currently the travis script retains "trusty" as the build environment, like it was before. If this one works on CI, a next iteration can update the recipes to be even more compliant with current zproject.

PLEASE DO NOT MERGE before this ends up green in CI ;)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,4 @@
 # This is a skeleton created by zproject.
 # You can add hand-written code here.
-
 # disables auto CRLF conversion for all files; create the file correctly and it will be allright
 * -text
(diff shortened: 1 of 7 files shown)


jimklimov:
The "check_zproject" recipe fails because OBS is busy and has not packaged the recently updated version of zproject yet, so templates seem to do differ as far as the test is concerned.

One way out of this is to check against the real upstream zproject from source code, as it tends to evolve regularly and lags due to packaging are undesired.

PR #1988, opened 2019-02-02
 Problem: CZMQ doesn't have an http server 

somdoron:
Problem: Using http server libraries together with czmq is hard.
You have to integrate czmq event loop with the the library event loop, which is not an easy task.

Solution: Wrap an existing library and provide CZMQ style api.


project.xml
```diff
@@ -15,6 +15,10 @@
     <use project = "systemd" optional = "1" implied = "1" min_major = "200" />
     <use project = "lz4" optional = "1" implied = "1" />
     <use project = "libcurl" optional = "1" implied = "1" min_major = "7" min_minor = "28" />
+    <use project = "libmicrohttpd"
+         tarball = "http://ftp.gnu.org/gnu/libmicrohttpd/libmicrohttpd-0.9.61.tar.gz"
+         implied = "1" optional = "1" min_major = "0" min_minor = "9" min_patch = "61"  />
```
bluca: @somdoron any reason for this min version? It's what breaks using the package - it's not available anywhere yet

somdoron: Yes, but we might be able to reduce it, let me check, because I'm using some method that is only available recently

bluca: Here you can find what's available on Ubuntu:

https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libmicrohttpd-dev&searchon=names

bluca: Ah ok that's even better - there's another change to do for backward compatibility:

```
--- a/src/zhttp_server.c
+++ b/src/zhttp_server.c
@@ -108,7 +108,11 @@ server_actor_new (zsock_t *pipe, zhttp_server_options_t *options)
     assert (self);
 
     self->daemon = MHD_start_daemon (
+#ifdef MHD_ALLOW_SUSPEND_RESUME
             MHD_ALLOW_SUSPEND_RESUME, // TODO: use epoll on linux
+#else
+            MHD_USE_SUSPEND_RESUME, // TODO: use epoll on linux
+#endif
             (uint16_t ) zhttp_server_options_port (options),
             NULL, NULL,
             (MHD_AccessHandlerCallback) s_handle_request, self,
```

somdoron: its an enum (:

somdoron: but they have the same value, so I will use the old one for now

somdoron: https://travis-ci.org/somdoron/czmq

bluca: Ah, the upper case fooled me - it's easy to add an autoconf/cmake check if you want

somdoron: We can just have a version check, do you know which version it was introduced?

bluca: 0.9.53 https://github.com/Karlson2k/libmicrohttpd/commit/959d55ec5eea5b5d68fed524d1b48184c4763620

somdoron: yes, just found it as well.
Made the change, CI is building, will make the pull request.
I will be in the airplane real soon, so hopefully everything will pass.

bluca: ok, safe travels!

src/zhttp_server.c
```diff
@@ -0,0 +1,512 @@
...
+}
+
+static server_actor_t *
+server_actor_new (zsock_t *pipe, zhttp_server_options_t *options)
+{
+    server_actor_t *self = (server_actor_t *) zmalloc (sizeof (server_actor_t));
+    assert (self);
+
+    self->daemon = MHD_start_daemon (
+            MHD_ALLOW_SUSPEND_RESUME, // TODO: use epoll on linux
```
bluca: @somdoron this was introduced in 0.9.34 by https://github.com/scottjg/libmicrohttpd/commit/b19e9c0cf0a3aef3794b5994288b0478b20e9490 - and SUSE until Leap 15 has 0.9.33. Can this flag be omitted if not available, or will it break things?

somdoron: I think we just need to have another version flag, if less than 0.9.34 we should use MHD_USE_PIPE_FOR_SHUTDOWN instead.

bluca: ok, can send a PR in a minute

somdoron: Cool

src/zhttp_server.c
```diff
@@ -0,0 +1,512 @@
...
+
+        free (self);
+    }
+
+    *self_p = NULL;
+}
+
+static void *
+s_insert_connection (struct MHD_Connection *connection) {
+    MHD_suspend_connection (connection);
```
bluca: this API and the next were also introduced in 0.9.34: https://github.com/scottjg/libmicrohttpd/commit/64a2c30ad3c3476bf16f0b2b579e040c6b65922a is there an alternative?
I can backport for the OBS projects if needed

bluca: @somdoron

somdoron: I'm not sure it will work...
Actually I think it will work, but will cause very high CPU as microhttpd will continue call the handler function until response is provided.

bluca: I think that as long a we document that, it's fine. It's an optional feature and we can recommend to use at least 0.9.34 if the intention is to deploy in production. What do you think?

somdoron: Sounds good to me. 
I'm not 100% sure it will work, but I think so.
So just skip the suspend and resume functions if not available.

bluca: Ok, will send a PR later

PR #2000, opened 2019-02-24
Problem: Appveyor does not include czmq's dll in artifacts

sappo:
Solution: Use the correct dll name in the zip command


bluca:
uh, so we can publish windows binaries from appveyor?

sappo:
I guess so. In fact each build produces an zip files which can be
downloaded on the artifacts tab in appveyor.

Luca Boccassi <notifications@github.com> schrieb am So., 24. Feb. 2019,
20:21:

bluca:
interesting, we could do release builds from it then I guess - @sigiesec would something like that be useful for libzmq?

PR #2006, opened 2019-05-16
Problem: zproject out of date

sphaero:
Solution: update from zproject

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN sudo make install
 RUN sudo ldconfig
 
 WORKDIR /home/zmq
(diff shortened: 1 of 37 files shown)


jimklimov:
It seems one new file was generated and was missed in the PR?

From https://travis-ci.org/zeromq/czmq/jobs/533238572 :
````
+git --version
git version 2.21.0
++git --no-pager diff -w
+[[ -n '' ]]
++git status -s
+[[ -n ?? bindings/python/_project.py ]]
+git status -s
?? bindings/python/_project.py
+echo 'zproject generated new files!'
zproject generated new files!
+exit 1
The command "./ci_build.sh" exited with 1.
````

PR #2012, opened 2019-06-10
Problem: Failed to build czmq on Windows via CMake

### LINKED ISSUE #1972: Failed to build czmq on Windows via CMake
Hey, guys! I am back again.

I am trying to add `czmq` to `vcpkg` these days(https://github.com/Microsoft/vcpkg/pull/4979). When building `czmq` with CMake, I encountered several problems.

1. <del>`CMAKE_CURRENT_SOURCE_DIR` isn't in `CMAKE_MODULE_PATH`.</del>

   <del>https://github.com/zeromq/czmq/blob/master/CMakeLists.txt#L19</del>
  
   <del>so `Findlibzmq.cmake`, `Findlibsodium.cmake`, .etc. may not be called.</del>

2. <del>When built as a static library,  libzmq will set an additional compiler flag `-DZMQ_STATIC` through `set_target_properties`:</del>

   <del>https://github.com/zeromq/libzmq/blob/master/CMakeLists.txt#L1060</del>

   <del>so if we want to build `czmq` as a static library on Windows, we should define `ZMQ_STATIC` manually(Maybe we can define it in `Findlibzmq.cmake`). Otherwise, the build will fail: https://github.com/zeromq/libzmq/issues/3318  </del>

3. `Findlibzmq.cmake` is not be able to find `libzmq` on Windows.

   libzmq dll/lib built with MSVC is named using the Boost convention(something like `libzmq-mt-4_3_1.lib`, `libzmq-mt-4_3_1.dll`). So `find_library (LIBZMQ_LIBRARIES NAMES zmq HINTS ${PC_LIBZMQ_LIBRARY_HINTS})` will not work.

    Here is a patch for solving this problem: https://bitbucket.org/ignitionrobotics/ign-transport/pull-requests/19/detec-zmq-under-windows-using-a-findzmq/diff

4. `Findlibsodium.cmake` will not found `libsodium` on Windows

   When built with MSVC, `libsodium` will generate a library named `libsodium.dll/lib`. In this case, `find_library(LIBSODIUM_LIBRARY NAMES sodium)` is not able to find the library. I found that `find_library(LIBSODIUM_LIBRARY NAMES sodium libsodium)` works.

   By the way, `Findlibcurl.cmake` has this problem too.

5. Maybe we can add some options like:
    
    <del> `WITH_LIBSODIUM`</del>
    `WITH_LIBCURL`
    `WITH_LZ4`
    `WITH_UUID`
    <del>`BUILD_TESTS` (For example: https://github.com/GoogleCloudPlatform/google-cloud-cpp/issues/1617)</del>
    `BUILD_TOOLS` (zmakecret)

myd7349:
Fix #1972 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 ########################################################################
 # Project setup
 ########################################################################
-cmake_minimum_required(VERSION 2.8.8)
+cmake_minimum_required(VERSION 2.8.12)
 project(czmq)
 enable_language(C)
 enable_testing()
@@ -138,7 +138,8 @@ ENDIF (LIBZMQ_FOUND)
 # UUID dependency
 ########################################################################
 find_package(uuid)
-IF (UUID_FOUND)
+option(CZMQ_WITH_UUID "Build czmq with uuid" ${UUID_FOUND})
+IF (CZMQ_WITH_UUID AND UUID_FOUND)
     include_directories(${UUID_INCLUDE_DIRS})
     list(APPEND MORE_LIBRARIES ${UUID_LIBRARIES})
   IF (PC_UUID_FOUND)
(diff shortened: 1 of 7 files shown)


sappo:
Neat. Great Readme section :+1:

PR #2028, opened 2019-10-28
Problem: zpoller docs isn't mentioning file handles

sphaero:
Problem: zpoller api doc isn't mentioning file handles

Solution: update api docs


bluca:
can you also regenerate please? otherwise ci will fail

sphaero:
sorry, forgot

PR #2033, opened 2019-12-14
Problem: czmq cannot be used as a subproject

madebr:
Solution: Use PROJECT_SOURCE_DIR and PROJECT_BINARY_DIR instead of
CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ endif()
 
 file(REMOVE "${SOURCE_DIR}/src/platform.h")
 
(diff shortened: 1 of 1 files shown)


bluca:
Thanks for the PR, but these are autogenerated, so please change at the source: https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl

madebr:
Oops, sorry!
Luckily, the change is easy to transfer.

PR #2037, opened 2020-01-17
Problem: openSUSE packaging requires SPDX formatted license

miska:
Solution: Change license name formating to SPDX format in spec and project
metadata (so regenerating spec would keep the change).


bluca:
Could you please send the same to https://github.com/zeromq/zproject/blob/master/zproject_redhat.gsl ? Otherwise it will get overwritten as it's a generated file

miska:
That should propagate value from the xml in this project which I changed as well.

bluca:
Ah yes, of course

PR #2039, opened 2020-01-19
Problem: NEWS out of date with recent changes

bluca:
Solution: update it


bluca:
@somdoron do we need to do anything in CZMQ for websockets?

somdoron:
add the wss socket options

bluca:
Ok, I'll add them

PR #2046, opened 2020-01-30
Problem: API out of date with new functions zframe_print_n and zmsg_print_n

bluca:
Add them as DRAFT, regen and update NEWS


pjaitken:
Thanks @bluca.

PR #2064, opened 2020-03-06
Enable ipv4 multicast packet types when using zbeacon

Dysl3xik:
Problem: ipv4 multicast packet types when using zbeacon are not implemented - but some notes about it already exist in the code

Solution: Enable ipv4 multicast packet types when using zbeacon

Open Items:
I could not find the files used to gen czmq_selftest or czmq_private_selftest, thefore I could not update the auto code generator files for these. Do they exist?

I intentionally did not modify any behavior for ipv6 with this change zsys_udp_new() has the code to configure the socket differently for IPv6 but its not enabled.

I have not tested this code on any OS aside from windows. There I tested it going through atleast 1 router and verified TTL and addresses were set correctly with wire shark. Additionally I copied the existing beacon test and added the calls to configure it with multicast. These tests pass as well.

Please advise how we can best move this code forward and hopefully have it added into official codebase.

### DIFF
--- a/src/zbeacon.c
+++ b/src/zbeacon.c
@@ -89,6 +89,8 @@ s_self_prepare_udp (self_t *self)
     if (self->udpsock_send != INVALID_SOCKET)
         zsys_udp_close (self->udpsock_send);
 
+    bool enable_multicast = zsys_ipv4_mcast_address () != NULL;
+
     self->hostname [0] = 0;
     //  For IPv6 we need two sockets. At least on Linux, IPv6 multicast packets
     //  are NOT received despite joining the group and setting the interface
@@ -101,12 +103,12 @@ s_self_prepare_udp (self_t *self)
     //  As a workaround, use a different socket to send packets. So the socket
     //  that receives can be bound to in6_addrany, and the socket that sends
     //  can be bound to the actual intended host address.
-    self->udpsock = zsys_udp_new (false);
+    self->udpsock = zsys_udp_new (enable_multicast);
     if (self->udpsock == INVALID_SOCKET) {
         self->udpsock_send = INVALID_SOCKET;
         return;
     }
-    self->udpsock_send = zsys_udp_new (false);
+    self->udpsock_send = zsys_udp_new (enable_multicast);
     if (self->udpsock_send == INVALID_SOCKET) {
         zsys_udp_close (self->udpsock);
         self->udpsock = INVALID_SOCKET;
@@ -146,6 +148,11 @@ s_self_prepare_udp (self_t *self)
                     &send_to);
             assert (rc == 0);
         }
+        else if (enable_multicast) {
+            rc = getaddrinfo (zsys_ipv4_mcast_address(), self->port_nbr, &hint,
+                    &send_to);
+            assert (rc == 0);
+        }
         else {
             rc = getaddrinfo ("255.255.255.255", self->port_nbr, &hint,
                     &send_to);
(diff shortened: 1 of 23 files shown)


src/zbeacon.c
```diff
@@ -146,6 +148,11 @@ s_self_prepare_udp (self_t *self)
                     &send_to);
             assert (rc == 0);
         }
+        if (enable_multicast) {
```
bluca: This doesn't look right - ipv6 will always be disabled regardless of the settings, as it will be overwritten by either v4 multicast or v4 broadcast

bluca: iow: there's an "else" missing

bluca:
> I could not find the files used to gen czmq_selftest or czmq_private_selftest, thefore I could not update the auto code generator files for these. Do they exist?

It's zproject via gsl
https://github.com/zeromq/zproject
https://github.com/zeromq/gsl

sappo:
@Dysl3xik thanks for this contribution. Its okay with our contribution guidelines if you only tested on one OS. You can check travis to see if it build on linux on macos if you like though. Currently the build fails on travis but works on my linux machine so that's not a blocker for me.

For further contributions please use the Problem/Solution statement in your commit messages.

Don't worry about the auto-generated files I'll take care of it once the PR is merged.

src/zsys.c
```diff
@@ -1071,6 +1078,25 @@ zsys_udp_new (bool routable)
                     (char *) &on, sizeof (on)) == SOCKET_ERROR)
         zsys_socket_error ("setsockopt (SO_REUSEPORT)");
 #endif
+
+    //  Only set TLL for IPv4
+    if (routable) {
```
bluca: IIRC setting the TTL/HOPS might be blocked and might fail. This should be attempted only if it's > 1 to avoid erroring out for no reason.

src/zbeacon.c
```diff
@@ -654,3 +711,135 @@ zbeacon_test (bool verbose)
     //  @end
     printf ("OK\n");
 }
+
+
+//  --------------------------------------------------------------------------
+//  Selftest
+
+void zbeacon_ipv4_mcast_test (bool verbose)
```
bluca: Don't add a new function, fold it in the existing one. Also, you need to do a check if setting v4 multicast addresses is allowed on the platform at all - it might not, and that's not a reason to fail the test. For example, all CI runs are broken due to this right now. This needs to be fixed before it can be merged.

Dysl3xik: Is it OK if I just chain this off the existing function? i.e. call this at the end of the existing test?

bluca: yes

Dysl3xik: For your other request to check if setting the address is allowed, is there a special function for this? Looking around on the internet everyone is just suggesting running ifconfig or netstat -g.

I am not seeing anything programatically showing how to check this.

bluca: What we usually do in unit tests is simply trying to set it, and fail gracefully (hiding the error) if it doesn't work

src/zbeacon.c
```diff
@@ -585,17 +585,18 @@ zbeacon (zsock_t *pipe, void *args)
...
 }
 
-
 //  --------------------------------------------------------------------------
-//  Selftest
+//  Selftest Multicast
 
-void
-zbeacon_test (bool verbose)
+void zbeacon_ipv4_mcast_test (bool verbose)
```
bluca: This needs to be static void, otherwise it will be exported as a public symbol

src/zbeacon.c
```diff
@@ -705,8 +706,11 @@ zbeacon_test (bool verbose)
     zactor_destroy (&node2);
     zactor_destroy (&node3);
 
-#if defined (__WINDOWS__)
-    zsys_shutdown();
+    //Unset multicast
+    zsys_set_ipv4_mcast_address (NULL);
+
+#if defined(__WINDOWS__)
```
bluca: Remove this, and call it from before in the _test function. We should not run shutdown multiple times.

src/zbeacon.c
Dysl3xik: Man you are fast! Which is quite nice when trying to wrap up stuff like this 😄 

As for trying to fail gracefully I am struggling with that. The fact that the beacon is run on a pipe/thread with only one way signaling makes it very hard to trickle any status stuff back out also all the function calls like s_self_prepare_udp return void so its difficult to even call those directly and check for failure that way.

I can try to make a socket directly using zsys_udp_new (enable_multicast) and set the flags on that socket, but that isn't really testing anything it would just cause it to abort running the "real" code. Also this would duplicate a huge chunk of the logic in s_self_prepare_udp as most of that logic is required to get to a state where you are ready to call set multicast.

Do you have any suggestions on a better/ideal approach?

bluca: Copy a shortened implementation of the bsd socket creation with all the multicast flags, and run it before the test. If it works, then the test can go ahead, otherwise it can be skipped. This is how we test for ipv6 support for example.

Dysl3xik:
Failures now seem limited to "implicit declaration of function 'zsys_set_ipv4_mcast_address'" which I do not understand as this function is defined in zsys.h and the api file....

Additionally the windows tests are randomly failing on poller. **I see this behavior randomly on my box as well**. I also see random failures of the ztimer test on my PC also randomly.....

bluca:
After updating the API files, the project needs to be regenerated with gsl, like every other zproject repo. I've done it for you. You should really get that set up.

sappo:
@Dysl3xik I build a Docker image for zproject to re-generate generated files in czmq.

It would be great if you could try it on windows and report if it works for you:
https://github.com/zeromq/czmq#code-generation

PR #2073, opened 2020-03-20
Using ZeroMQ cmake package config file if available

Mystfit:
```
Problem: ZeroMQ provides a cmake package config file that we can use

Solution: If available use the ZMQ package config file, otherwise fallback to FindlibZMQ.cmake
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,19 +120,30 @@ set(OPTIONAL_LIBRARIES_STATIC)
 ########################################################################
 # LIBZMQ dependency
 ########################################################################
-find_package(libzmq REQUIRED)
-IF (LIBZMQ_FOUND)
-    include_directories(${LIBZMQ_INCLUDE_DIRS})
-    list(APPEND MORE_LIBRARIES ${LIBZMQ_LIBRARIES})
-  IF (PC_LIBZMQ_FOUND)
-      set(pkg_config_names_private "${pkg_config_names_private} libzmq")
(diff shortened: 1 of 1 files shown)


bluca:
needs to be changed in here https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl otherwise it will be overwritten on regen

sappo:
What is the advantage of this approach?

sappo:
Windows builds on appveyor are failing!

Mystfit:
> What is the advantage of this approach?

My thought was that since ZeroMQ is generating a cmake package file that contains all the necessary exported targets and transitive information (compile definitions and include directories etc) that it would easier to use that instead of FindlibZMQ.cmake which has to be maintained in parallel.

Mystfit:
> Windows builds on appveyor are failing!

I'll look into it. Either the include directories aren't coming across or some other information. I'm going to redo this PR using zproject anyway since I went about adding these changes incorrectly, my apologies.

sappo:
If you need help https://gitter.im/zeromq/zproject.

Byron Mallett <notifications@github.com> schrieb am Sa., 21. März 2020,
03:52:

sappo:
> > Windows builds on appveyor are failing!
> 
> I'll look into it. Either the include directories aren't coming across or some other information. I'm going to redo this PR using zproject anyway since I went about adding these changes incorrectly, my apologies.

No apologies required! Your approach to introduce those changes was perfectly valid. Thanks for your effort so far :+1:

PR #2076, opened 2020-04-19
problem: no support for disconnect and hello msg socket options

somdoron:

### DIFF
--- a/api/zsock_option.api
+++ b/api/zsock_option.api
@@ -16,6 +16,18 @@
     <argument name = "only first subscribe" type = "integer" />
 </method>
 
+<method name = "set hello msg" polymorphic = "1">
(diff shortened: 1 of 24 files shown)


api/zsock_option.api
```diff
@@ -16,6 +16,18 @@
     <argument name = "only first subscribe" type = "integer" />
 </method>
 
+<method name = "set hello msg" polymorphic = "1">
```
bluca, line 19: these both need to be draft

bluca: actually never mind, it's not manually added

bluca:
you need to regen, for the bindings

somdoron:
regenerated

PR #2082, opened 2020-05-18
FIX #2081 - ZPL: zconfig_chunk_load() fails with 'large' property value

### LINKED ISSUE #2081: ZPL: zconfig_chunk_load() fails with 'large' property value
Description;
Giving a large string value  for a property of a zpl, the serialize/parse process fails.
The size of the value should be at least 1024 chars.

Code (test) to reproduce:
```
void main (int, const *char[])
{
    zconfig_t *root = zconfig_new ("root", NULL);
    zconfig_t *item = zconfig_new ("item", root);
    assert(root && item);
    size_t sz = 2048; // === large size, at least 1024
    char *value = (char*) zmalloc (sz + 1);
    assert (value);
    memset (value, 'a', sz); 
    value[sz] = '\0';
    zconfig_set_value (item, "%s", value);

    // serialize root in string, str parse root2 from string
    zconfig_t *root2 = NULL;
    {
            char *saved = zconfig_str_save (root);
            assert (saved);
            root2 = zconfig_str_load (saved);
            free (saved);
    }

     printf("size: %d, root2 %s\n", sz, (root2 ? "OK" : "KO"));
     assert (root2); // === assertion fails, root2 is NULL

     const char *value2 = zconfig_get (root2, "item", NULL);
     assert (value2);
     assert (streq (value, value2));

     free (value);
     zconfig_destroy (&root2);
     zconfig_destroy (&root);
}
```

Fix:
avoid abrupt truncate (1024 bytes) of chunk buffer in zconfig_chunk_load ()
https://github.com/zeromq/czmq/blob/master/src/zconfig.c#L679

jeffdegott:
https://github.com/zeromq/czmq/issues/2081

### DIFF
--- a/src/zconfig.c
+++ b/src/zconfig.c
@@ -665,6 +665,9 @@ zconfig_chunk_load (zchunk_t *chunk)
     char *data_ptr = (char *) zchunk_data (chunk);
     size_t remaining = zchunk_size (chunk);
 
+    char cur_line_p [1024 + 1]; // predefined line buffer (to limit zmalloc() calls)
+    char *cur_line = (char *) NULL;
+
     while (remaining) {
         //  Copy stuff into cur_line; not fastest but safest option
         //  since chunk may not be null terminated, etc.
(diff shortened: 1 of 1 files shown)


jeffdegott:
travis job 3486.19 failure on:
ERROR: JAVA_HOME is set to an invalid directory: C:\Program Files\OpenJDK\jdk-13.0.2

sphaero:
Why hasn't this been merged yet? The CI fails are not related to this change.

PR #2087, opened 2020-06-13
Problem: logsender destroyed before last log message

sappo:
Solution: Destroy logsender only just before the zmq context is shutdown

@bluca can you still find any race conditions?

### DIFF
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -69,6 +69,7 @@ s_handler_fn_shim (DWORD ctrltype)
 //  ZeroMQ context for this process
 static void *s_process_ctx = NULL;
 static bool s_initialized = false;
+static bool s_shutting_down = false;
 
 #ifndef S_DEFAULT_ZSYS_FILE_STABLE_AGE_MSEC
 // This is a private tunable that is likely to be replaced or tweaked later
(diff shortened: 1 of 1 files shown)


src/zsys.c
```diff
@@ -347,6 +347,8 @@ zsys_shutdown (void)
 
     ZMUTEX_DESTROY (s_mutex);
 
+    s_initialized = false;
```
bluca, line 350: Same as before, the mutex might be destroyed while still in use - this should be the first thing that happens, as before, to avoid any potential race

sappo: Sorry, I don't get it :(

bluca: s_initialized, being a boolean and thus below word size, is implicitly atomic. On top of that, most architectures have cache coherency guarantees for <= word size stores.
So even if you have multiple threads entering zsys_shutdown, if reading and changing s_initialized is the first thing that happens, you are pretty much guaranteed that only one will make it through, and the rest will exit immediately. At least on architectures people actually use.

By moving this store later on, this property is lost, and now it's possible that multiple threads can start processing zsys_shutdown at the same time, which might result in undefined behaviour due to the mutex, which gets destroyed while others might be blocked on it.

sappo: @bluca thanks for explaining. What's your take on this solution?

bluca: Looks good, thanks!

PR #2091, opened 2020-06-28
Problem: OBS build on sid fails

bluca:
Solution: regen from zproject to change the version format to non-native as a hack,
to match OBS' debstransform usage of 1.0 format


sappo:
There's something wrong with the android build. Can you exclude that from the regen. I'll investigate why this changed.

bluca:
> There's something wrong with the android build. Can you exclude that from the regen. I'll investigate why this changed.

ok, done

PR #2092, opened 2020-07-17
Add new zudp class for UDP sockets

sphaero:
Problem: zsys only has limited support for UDP sockets and mentions we need to move things to a zudp class
Solution: create a new zudp class

This is an initial class which supports unicast sockets as well. The class is in draft state so we can still work things out. The broadcast socket still needs work. IPv6 multicast on OSX also needs attention but I need to get to one of those machines first. The test skips ipv6 multicast on OSX. 

This PR also adds a method to zsys to test whether ipv6 is available. This is ported from libzmq.

Review and feedback welcomed


somdoron:
Are you familiar with ZMQ_DGRAM? It's a libzmq UDP socket

sphaero:
I'm familiar with the RADIO/DISH sockets but I thought these were using a UDP transport and not raw UDP sockets? I have to use cases for UDP sockets. One is zbeacon. The other is communicating with other applications that already support UDP. For example many applications support OSC (Open Sound Control) which is just some serialisation on top of UDP.

Are you saying I can use ZMQ_DGRAM for this? I suppose you mean this: http://api.zeromq.org/master:zmq-udp

somdoron:
Checkout https://github.com/zeromq/libzmq/blob/master/tests/test_dgram.cpp

DGRAM is a socket type, similar to STREAM socket type. It is just native UDP socket. When calling send you can specify the destination ip address (like sendto).

bluca:
Yeah I think it would make sense to enchance the ZMQ_DGRAM to add any missing features - I think the UDP transport already supports multicast, I'm not sure about the IPv6 status

sphaero:
I need to read into this and test whether this can replace it. Would be great!

Is this already exposed in czmq? Or what would we need to do so?

bluca:
Add a usual wrapper in the zsock class, I think - that should be enough

sphaero:
Ok, I'm going to test this. But just to be sure; what's the difference between the ZMQ_DGRAM socket and a regular UDP socket? If any?

somdoron:
None, first frame is the IP and port, second frame is the message.

sphaero:
But I assume that's not on the wire, right? The receive is adding that?

somdoron:
Yes, on the wire it plain UDP message, no envelope or frame

sphaero:
I've looked into the ZMQ_DGRAM sockets and it kind of works. I have modified zsock to include the ZMQ_DGRAM sockets. Haven't tested multicast yet but ipv4 unicast seems fine, ipv6 unicast not. Ipv6 does bind succesfully and it shows in the netstat table:
```
$ sudo netstat -anp | grep czmq
udp6       0      0 :::7777                 :::*                                268277/czmq_selftes 
udp6       0      0 :::58838                :::*                                268277/czmq_selftes 
```
However the send does nothing. Any suggestions?

Here's the test code run by czmq_selftest
```
#ifdef ZMQ_DGRAM
    // ZMQ_DGRAM tests
    zsock_t* dgramr = zsock_new_dgram ("udp://*:7777");
    assert (dgramr);
    zsock_t* dgrams = zsock_new_dgram ("udp://*:*");
    assert (dgrams);
    // perhaps sleep a little for sockets to setup?

    rc = zstr_sendm( dgrams, "127.0.0.1:7777" );
    assert (rc == 0);
    rc = zstr_send (dgrams, "HELLO");
    assert (rc == 0);

    char *dmessage, *addr;

    zmsg_t *dmsg = zmsg_recv( dgramr );
    assert (dmsg);
    addr = zmsg_popstr (dmsg);
    dmessage = zmsg_popstr (dmsg);
    assert (streq(dmessage, "HELLO"));
    zmsg_destroy ( &dmsg );
    zsock_destroy (&dgrams);
    zsock_destroy (&dgramr);
    zstr_free (&dmessage);
    zstr_free (&addr);
    zstr_free (&message);

    // ipv6
    zsys_set_ipv6(1);
    zsock_t* dgramr6 = zsock_new_dgram ("udp://*:7777");
    assert (dgramr6);
    zsock_t* dgrams6 = zsock_new_dgram ("udp://*:*");
    assert (dgrams6);
    // perhaps sleep a little for sockets to setup?
    zclock_sleep(100);

    rc = zstr_sendm( dgrams6, "::1:7777" );
    assert (rc == 0);
    rc = zstr_send (dgrams6, "HELLO");
    assert (rc == 0);

    char *dmessage6, *addr6;
    zmsg_t *dmsg6 = zmsg_recv( dgramr6 );
    assert (dmsg6);
    addr6 = zmsg_popstr (dmsg6);
    dmessage6 = zmsg_popstr(dmsg6);
    assert (streq(dmessage6, "HELLO"));
    zmsg_destroy( &dmsg6 );
    zsock_destroy (&dgrams6);
    zsock_destroy (&dgramr6);

    zstr_free (&dmessage6);
    zstr_free (&addr6);
    zstr_free (&dmessage6);

#endif
```

bluca:
yeah the udp engine needs some improvements to handle ipv6 - but shouldn't be too much work

sphaero:
perhaps I have to rise up to the challenge of the libzmq core as I agree this is better handled through libzmq. A PR is not the right place to track this, right? I'll make a note of the following todo's:
* create a PR to expose new sockets in czmq through zsock + tests
* create an issue for libzmq about ipv6 support in ZMQ_DGRAM (is that the best place to track this?)

@bluca if you have any pointers where the work resides in libzmq I can have a look. Probably in udp_engine.cpp?

somdoron:
sockaddr_to_msg:
https://github.com/zeromq/libzmq/blob/master/src/udp_engine.cpp#L364

resolve_raw_address:
https://github.com/zeromq/libzmq/blob/master/src/udp_engine.cpp#L392

sphaero:
Reported here:
https://github.com/zeromq/libzmq/issues/3993

Closing this PR

PR #2094, opened 2020-07-19
Problem: ZMQ_DGRAM sockets are not exposed

sphaero:
Solution: wrap ZMQ_DGRAM in the zsock class and add some tests.


include/zsock.h
```diff
@@ -1270,6 +1270,20 @@ CZMQ_EXPORT zsock_t *
 CZMQ_EXPORT zsock_t *
     zsock_new_scatter (const char *endpoint);
 
+//  *** Draft method, for development use, may change without warning ***
+//  Create a DGRAM (UDP) socket. Default action is bind.
+//  The endpoint is a string consisting of a
+//  'transport'`://` followed by an 'address'. As this is
+//  a UDP socket the 'transport' has to be 'udp'. The
+//  'address' specifies the ip address and port to
+//  bind or connect to. For example:  udp://127.0.0.1:1234
```
somdoron, line 1279: Only bind

include/zsock.h
```diff
@@ -1270,6 +1270,20 @@ CZMQ_EXPORT zsock_t *
...
+//  Note: a DGRAM socket can only connect to a RADIO socket!
```
somdoron, line 1280: DGRAM can only connect to DGRAM or raw UDP

sphaero: see latest commits

PR #2099, opened 2020-07-28
Adding OSC support to cmzq

sphaero:
This PR is not fixing anything but rather adding OSC support as a new feature.

[Open Sound Control](http://opensoundcontrol.org/) (OSC) is a serialisation format (and usually transported over UDP) which is supported by many applications and appliances. It was developed at CNMAT and by now a de facto protocol for networking sound synthesizers, computers, and other multimedia devices for purposes such as musical performance or show control. It is also often used for rapid prototyping purposes due to the support by many applications and frameworks in this field. The name can be bit misleading as it is much wider used than only sound and music field. The OSC format is perfectly usable with ZeroMQ sockets. With ZeroMQ's DGRAM sockets it is also possible to use ZeroMQ to send and receive OSC messages which can be understood by any device supporting OSC. 

An OSC message is a quite simple format consisting of an address string and a format string describing the data. The format string type tags are as follows:
* i - 32bit integer
* h - 64bit integer
* f - 32bit floating point number (IEEE)
* d - 64bit (double) floating point number
* s - string (NULL terminated)
* t = timetag: an OSC timetag in NTP format (uint64_t)
* S - symbol
* c - char
* m - 4 byte midi packet (8 digits hexadecimal)
* T - TRUE (no value required)
* F - FALSE (no value required)
* N - NIL (no value required)
* I - Impulse (for triggers) or INFINITUM (no value required)
* b - binary blob

I'm not sure if this is wanted in the czmq framework as there are no other encoders/decoders but that can be discussed here. If wanted the PR can also be squashed to a cleaner change.


bluca:
Nice - Windows builds are unhappy tho

sphaero:
Yeah I noticed, somehow it can't find the `htonll` methods which are inlines of winsock2.h. I need to look into that.

sphaero:
Windows builds and tests fine. However it errors on java:
```
Successfully started process 'command 'javac''
javac: file not found: src\main\java\org\zeromq\czmq\Zosc.java
```
I've never done anything with the java bindings. Is there anything I need to add to make it run successful?

bluca:
@sappo any idea about java?

sphaero:
I found that the jni binding files are ignored by some .gitignore file. I force added them and now the build succeeds, apart from a segfault on zpoller or zsock now and then. I can't reproduce these segfaults but they happen also without this addition.

sphaero:
Should I add these generated files as well btw?
```
	bindings/python_cffi/czmq_cffi/Zosc.py
	bindings/qml/src/QmlZosc.cpp
	bindings/qml/src/QmlZosc.h
	bindings/qt/src/qzosc.cpp
	bindings/qt/src/qzosc.h
	bindings/ruby/lib/czmq/ffi/zosc.rb
```

bluca:
Probably yes - although I'm not familiar with them, it won't hurt.

While you are there, could you also tidy up the branch history? Alternatively I can do a squash merge

taotetek:
Hey @sphaero I just wanted to say this is going to ruin my life as far as spending time on anything else, I have a Rebel Tech open sound control module in my modular synth haha.

sphaero:
I've squashed the history to 2 commits. @taotetek I'm sorry man. I was already hit by this virus, I guess it will spread through me. At least I hope you can inspire the whole community :grinning:

bluca:
still some CI failures:

src/zosc.c: In function 'zosc_frommem':
src/zosc.c:98:12: error: variable 'address_end' set but not used [-Werror=unused-but-set-variable]
     size_t address_end = 0;
            ^

sphaero:
Yes, missed that one, sorry. Also I have no access to an OSX machine so OSX is uncertain for the time. CI tests seem OK though.

PR #2104, opened 2020-08-12
Problem: useless print message in zgossip

stvales:
Solution: remove useless print message

### DIFF
--- a/src/zgossip.c
+++ b/src/zgossip.c
@@ -563,7 +563,6 @@ monitor_handler (zloop_t *loop, zsock_t *monitor_socket, void *argument)
     if (monitor) {
         zmsg_t *msg = zmsg_recv (monitor);
         if (msg) {
-            zmsg_print (msg);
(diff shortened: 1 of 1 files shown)


src/zgossip.c
```diff
@@ -563,7 +563,6 @@ monitor_handler (zloop_t *loop, zsock_t *monitor_socket, void *argument)
     if (monitor) {
         zmsg_t *msg = zmsg_recv (monitor);
         if (msg) {
-            zmsg_print (msg);
```
bluca, line 566: Why is it useless? Wouldn't it be better to hide it behind the verbose flag?

stvales: zgossip does not publicly support the VERBOSE command yet. I have commented some other code for the day it will support it. What makes things complicated is that zgossip relies on an automatically generated engine (zgossip_engine.inc) which supports verbose in its private structure but does not expose it.

PR #2108, opened 2020-09-13
Problem : various tests do not pass properly, compilation has many warnings in xcode

stvales:
Solution:
- solve bug in gossip monitor socket indexes
- adjust tests for zarmour and Z85 to work with empty strings
- add verbose messages for zsock tests
- adjust timer variable types to int64_t and adjust API to avoid compilation warnings on implicit types conversions

### DIFF
--- a/src/zgossip.c
+++ b/src/zgossip.c
@@ -563,7 +563,6 @@ monitor_handler (zloop_t *loop, zsock_t *monitor_socket, void *argument)
     if (monitor) {
         zmsg_t *msg = zmsg_recv (monitor);
         if (msg) {
-            zmsg_print (msg);
             char *event = zmsg_popstr (msg);
             assert (event);
             zframe_t *number = zmsg_pop (msg);


api/zclock.api
```diff
@@ -12,7 +12,7 @@
 
     <method name = "sleep" singleton = "1">
         Sleep for a number of milliseconds
-        <argument name = "msecs" type = "integer" />
+        <argument name = "msecs" type = "msecs" />
```
bluca, line 15: this is a stable api, signature cannot change

api/zpoller.api
```diff
@@ -55,7 +55,7 @@
         timeout expired, returns NULL. You can test the actual exit condition by 
         calling zpoller_expired () and zpoller_terminated (). The timeout is in 
         msec.
-        <argument name = "timeout" type = "integer" />
+        <argument name = "timeout" type = "msecs" />
```
bluca, line 58: this is a stable api, signature cannot change

stvales: There are inconsistencies in the API causing (a lot of) compilation warnings in clang. The proposed modifications do not introduce incompatibilities and enable to remove the warnings. Basically, we move from int to int64_t, which should always work.

bluca: Please see https://rfc.zeromq.org/spec/44/ - we do not do API changes after a class is declared stable, sorry

stvales: In this case, the proposed changes are to avoid actual possible data loss in the existing code. Warnings are generated for a reason. But I get it. The rules are the rules :-)

Issue #2113, opened 2020-09-17
zcertstore has no API to get the disk state from a custom loader


santupari:
Hello everyone,

We encountered 2 different issues listed below w.r.t CURVE authentication which uses zcertstore internally to store the client keys.
1. If there are multiple client certs with same timestamp or one of the cert has timestamp
      a. On client connection request certstore gets refreshed
      b. after that a new client cert is copied or valid cert is overwritten as the previous cert was invalid or corrupted, the certstore does not get refreshed and there after that connection will never be allowed as there is not change w.r.t timestamp, count and current size of the cert dir
2. If an application wants to load its custom disk loader (which loads all the certs on a change), then there is no API available to access the state store in zcertstore struct.

The above issues are observed on libczmq 4.2.0 version.
I have attached test code along with the patch for zcertstore.c and zcertstore.h.

Please let me know for any clarification.

[czmq_certstore_custom_loader.zip](https://github.com/zeromq/czmq/files/5235592/czmq_certstore_custom_loader.zip)

Thanks,
Santosh Bidaralli

bluca:
Please send a pull request with the patches rather than attachments

santupari:
Thank you @bluca. I will send the pull request at the earliest.

Yashashree-Jadhav:
Hi Everyone,

I am seeing the same issue as mentioned in  1st bullet here:

1. If there are multiple client certs with same timestamp or one of the cert has timestamp. On client connection request certstore
     gets refreshed after that a new client cert is copied or valid cert is overwritten as the previous cert was invalid or corrupted, 
    the certstore does not get refreshed and there after that connection will never be allowed as there is not change w.r.t 
    timestamp, count and current size of the cert dir

Before jumping to 4.2.1, I am curious how does the custom disk loader helps fix this issue.
I see in my case the default disk loader in zcertstore (s_disk_loader ) never makes into this if block:
**********************************************************
   if (dir
    && (state->modified != zdir_modified (dir)
    ||  state->count != zdir_count (dir)
    ||  state->cursize != (size_t) zdir_cursize (dir)))
    {
*********************************************************

santupari:
Yashashree,

You can create a custom disk loader that checks the hash of the directory (cert dir) every time a change is detected instead of count and size.

Thanks,
Santosh

PR #2114, opened 2020-09-18
zcertstore has no API to get the disk state from a custom loader #2113

santupari:
# Pull Request Notice

```
Problem: 
We encountered 2 different issues listed below w.r.t CURVE authentication which uses zcertstore internally to store the client keys.

If there are multiple client certs with same timestamp or one of the cert has timestamp
a. On client connection request certstore gets refreshed
b. after that a new client cert is copied or valid cert is overwritten as the previous cert was invalid or corrupted, the certstore does not get refreshed and there after that connection will never be allowed as there is not change w.r.t timestamp, count and current size of the cert dir
If an application wants to load its custom disk loader (which loads all the certs on a change), then there is no API available to access the state store in zcertstore struct.

Solution:
1. The custom disk loader can have its own way of determining the cert directory changes. In our case we created SHA1 of the directory and compared every time before refreshing the certstore (test code and patch is attached for reference)
[czmq_certstore_custom_loader.zip](https://github.com/zeromq/czmq/files/5242953/czmq_certstore_custom_loader.zip)

2. Added a new API zcertstore_get_state() that returns the user defined disk state to the caller

### DIFF
--- a/src/zcertstore.c
+++ b/src/zcertstore.c
@@ -267,6 +267,17 @@ zcertstore_print (zcertstore_t *self)
     }
 }
 
+//  --------------------------------------------------------------------------
+//  Return the state stored in certstore
+void *
+zcertstore_get_state (zcertstore_t *self)
+{
+    if (self)
+        return self->state;
+
+    return NULL;
+}
+
 
 //  --------------------------------------------------------------------------
 //  Selftest
--- a/include/zcertstore.h
+++ b/include/zcertstore.h
@@ -87,6 +87,11 @@ CZMQ_EXPORT void
 CZMQ_EXPORT zlistx_t *
     zcertstore_certs (zcertstore_t *self);
 
+//  --------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/zcertstore.c [hunk 1]
```diff
@@ -267,6 +267,17 @@ zcertstore_print (zcertstore_t *self)
...
 
+//  --------------------------------------------------------------------------
+//  Return the state stored in certstore
+void *
+zcertstore_get_state (zcertstore_t *self)
+{
+    if (self)
+        return self->state;
+
+    return NULL;
```
sappo, line 278: Our code style advocates to fail fast.
```c
assert (self);
return self->state;
```

sappo: It would be great if you could change this in another PR

sappo: Done

santupari: I saw your update. Thanks for fixing it. I will keep that in mind for future pull requests.

Issue #2038, opened 2020-01-17
ZMQ sockets (zsock_t) do not work in child process after fork


santupari:
Hi All,

We are observing an issue where in sockets created in child process do not work after forking.
Below is the sequence of code and I have also attached test code that reproduces issue and as well as working code:
1. Our design has a server which listens to multiple clients and responds
2. Clients uses wrapper API (example : test_tx_rx()) that exchanges messages with server
3. the issue observed when a client calls  test_tx_rx() from parent and then fork and if you call test_tx_rx() from child it fails to communicate with server
4. The fix is to call zsys_shutdown() from child to clean up all initialization done by parent and assume that ZSYS is already initialized. However, as part of this fix we also noticed that there is a mutex that should be re-initialized to avoid deadlock. Hence, we are also attaching a patch that defines new API for re-initializing the mutex.

Please let us know if the patch is valid or how to avoid deadlock in case if the parents threads are accessing the mutex at the same when child is clearing mutex in zsys_shutdown().

Thanks,
Santosh Bidaralli

[zsys_patch_wtih_test_code.zip](https://github.com/zeromq/czmq/files/4079091/zsys_patch_wtih_test_code.zip)

bluca:
Adding a new API is not good for usability.

A better fix is to record the PID on init, and check it on shutdown - if it's different, then a fork happened and the mutex can be reinitialised.

Could you please send a pull request for this change, after testing that it works for your use case? Thanks!

santupari:
Thank you @bluca!
I will make the suggested change in zsys_shutdown(). If my understanding is correct, defining a new static variable in zsys.c to store the parent pid should be sufficient to check it during the init and shutdown.
Please send me the instructions to submit a pull request.
Thanks,
Santosh Bidaralli

bluca:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests

santupari:
Thanks @bluca ! I found better way of handling this and with below changes there is no need for application to call zsys_shutdown(). Please review these changes and let me know if i can submit a pull request for the same.
 
--- orig/src/zsys.c	2020-01-27 16:07:47.089956960 -0600
+++ git/src/zsys.c	2020-01-27 16:09:34.408025720 -0600
@@ -108,6 +108,10 @@
 //  We keep a list of open sockets to report leaks to developers
 static zlist_t *s_sockref_list = NULL;
 
+//  We keep track of the pid of the process that creates the context
[...]

PR #2116, opened 2020-09-18
ZMQ sockets (zsock_t) do not work in child process after fork #2038

santupari:
We are observing an issue where in sockets created in child process do not work after forking.
Below is the sequence of code and I have also attached test code that reproduces issue and as well as working code:

Our design has a server which listens to multiple clients and responds
Clients uses wrapper API (example : test_tx_rx()) that exchanges messages with server
the issue observed when a client calls test_tx_rx() from parent and then fork and if you call test_tx_rx() from child it fails to communicate with server
The fix is to call zsys_shutdown() from child to clean up all initialization done by parent and assume that ZSYS is already initialized. However, as part of this fix we also noticed that there is a mutex that should be re-initialized to avoid deadlock. Hence, we are also attaching a patch that defines new API for re-initializing the mutex.

Please let us know if the patch is valid or how to avoid deadlock in case if the parents threads are accessing the mutex at the same time when child is clearing mutex in zsys_shutdown().

I am also attaching test code that can be used to reproduce the issue. This issue is no longer observed after the patch is applied.

Thanks,
Santosh Bidaralli
[zsys_patch_wtih_test_code.zip](https://github.com/zeromq/czmq/files/5247679/zsys_patch_wtih_test_code.zip)


bluca:
the windows build are broken

src/zsys.c
```diff
@@ -150,13 +153,46 @@ s_zsys_vprintf_hint (int hint, const char *format, va_list argptr);
 //  Not threadsafe, so call only from main thread. Safe to call multiple
 //  times. Returns global CZMQ context.
 
+// mutex for pthread_once to run the init function only once in a process
+static pthread_once_t init_all_mutex_var = PTHREAD_ONCE_INIT;
+
+// handler to initialize mutexes one time in multi threaded env
+void zsys_initialize_mutex() {
```
bluca, line 160: all these new functions should be static

santupari: i will send a new patch after changing these functions to static.

bluca:
the build still fails on windows, it needs to be fixed before this can be merged

santupari:
Thanks for the response. I will take a look at get back with a fix at the earliest.

bluca:
src/zsys.c is getting the executable bit set - likely a problem with your editor. Please remove it.

santupari:
I am sorry for the inconvenience. I have pushed the new file after removing executable bit.

PR #2121, opened 2020-10-02
Problem: Check for symbol AI_V4MAPPED based on incomplete platform list

sappo:
Solution: Instead of checking for platforms that don't have AI_V4MAPPED
defined - which may to change in the future - use autotools or cmake
to check for the symbol.

If found they'll declare the define HAVE_DECL_AI_V4MAPPED with value 1
otherwise 0.


sappo:
Fixes zeromq/zproject#1265
Fixes #2119

PR #2132, opened 2020-11-24
zosc_dup wasn't tested

sphaero:
Problem: zosc_dup isn't being tested

Solution: implement a test.


sphaero:
Also added a fix for when passing a NULL to zosc_dup

PR #2140, opened 2020-12-10
docs: fix simple typo, milticast -> multicast

timgates42:
There is a small typo in bindings/nodejs/README.md, bindings/python/czmq/_czmq_ctypes.py, bindings/python_cffi/czmq_cffi/Zsys.py, bindings/qml/src/QmlZsys.cpp, bindings/qml/src/QmlZsys.h, bindings/qt/src/qzsys.cpp, bindings/qt/src/qzsys.h, include/zsys.h.

Should read `multicast` rather than `milticast`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md

### DIFF
--- a/bindings/nodejs/README.md
+++ b/bindings/nodejs/README.md
@@ -4319,7 +4319,7 @@ Return IPv6 address to use for zbeacon reception, or "" if none was set.
 nothing my_zsys.setIpv6McastAddress (String)
 ```
 
-Set IPv6 milticast address to use for sending zbeacon messages. This needs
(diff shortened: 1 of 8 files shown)


bluca:
Thanks for the PR but all those files are generated, you need to fix it in api/zsys.api otherwise it will get overwritten again

sphaero:
root cause is here: https://github.com/zeromq/czmq/blob/fbe313cd2010bace7833fe52d419f82282343bd9/api/zsys.api#L479

PR #2111, opened 2020-09-15
Problem: zsock DGRAM multicast tests do not work in macOS

stvales:
Solution: add preprocessor condition to remove multicast test on Apple machines
Bonus: add comments for tests in verbose mode

PR #2162, opened 2021-02-04
Fix problems related to ZSYS_THREAD_NAME_PREFIX/ZSYS_THREAD_NAME_PREFIX_STR

skaes:
This patch addresses three problems related to ZSYS_THREAD_NAME_PREFIX / ZSYS_THREAD_NAME_PREFIX_STR:

1. Setting ZSYS_THREAD_NAME_PREFIX_STR is ignored (it reads ZSYS_THREAD_NAME_PREFIX when set)
2. ZSYS_THREAD_NAME_PREFIX is ignored (it uses the default for ZSYS_THREAD_NAME_PREFIX_STR instead)
3. Programs on MacOS immediately crash on startup (due to a strcpy(s,s), which is illegal)

I changed the minimal amount necessary to fix the problems, but I think this interface is confusing for users and does not really reflect the underlying libzmq implementation. I would much prefer a single environment variable ZSYS_THREAD_NAME_PREFIX where the implementation in czmq figures out which libzmq API to call, based on whether  the value specified is an integer or some other string.

### DIFF
--- a/src/zsock.c
+++ b/src/zsock.c
@@ -2376,6 +2376,7 @@ zsock_test (bool verbose)
     zstr_free (&addr);
     zstr_free (&message);
 
+#ifndef __APPLE__
     // ZMQ_DGRAM ipv4 multicast test
     zsock_t* mdgramr = zsock_new_dgram ("udp://225.25.25.25:7777");
     assert (mdgramr);
@@ -2389,6 +2390,7 @@ zsock_test (bool verbose)
 
     char *mdmessage, *maddr;
 
+    // this call blocks forever on MacOS
     zmsg_t *mdmsg = zmsg_recv( mdgramr );
     assert (mdmsg);
     maddr = zmsg_popstr (mdmsg);
@@ -2400,6 +2402,7 @@ zsock_test (bool verbose)
     zstr_free (&mdmessage);
     zstr_free (&maddr);
     zstr_free (&mdmessage);
+#endif
 
 //    // ipv6 (not supported yet)
 //    zsys_set_ipv6(1);
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -335,8 +335,6 @@ zsys_init (void)
 
     if (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"))
         zsys_set_thread_name_prefix_str (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"));
-    else
-        zsys_set_thread_name_prefix_str (s_thread_name_prefix_str);
 
     return s_process_ctx;
 }
--- a/src/zsys.c
+++ b/src/zsys.c
@@ -334,7 +334,7 @@ zsys_init (void)
         zsys_set_thread_name_prefix (s_thread_name_prefix);
 
     if (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"))
-        zsys_set_thread_name_prefix_str (getenv ("ZSYS_THREAD_NAME_PREFIX"));
+        zsys_set_thread_name_prefix_str (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"));
     else
         zsys_set_thread_name_prefix_str (s_thread_name_prefix_str);
 


src/zsys.c
```diff
@@ -334,9 +334,7 @@ zsys_init (void)
         zsys_set_thread_name_prefix (s_thread_name_prefix);
 
     if (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"))
-        zsys_set_thread_name_prefix_str (getenv ("ZSYS_THREAD_NAME_PREFIX"));
-    else
-        zsys_set_thread_name_prefix_str (s_thread_name_prefix_str);
+        zsys_set_thread_name_prefix_str (getenv ("ZSYS_THREAD_NAME_PREFIX_STR"));
```
bluca, line 337: Ops the wrong env var is clearly a copypasta, my bad - but iirc the "else" branch is required. Why is it being removed?

src/zsock.c
```diff
@@ -2376,6 +2376,7 @@ zsock_test (bool verbose)
     zstr_free (&addr);
     zstr_free (&message);
 
+#ifndef __APPLE__
```
bluca, line 2379: this smells a bit - any idea why multicast v4 wouldn't work on osx?

skaes: Of course it smells, but right now I can't run the tests unless this is commented out becauseI don't know how to re-configure the my Mac's networking to make it work. I'd be fine to keep this as a git stash to be able to work, but then everyone else will run into the same problem and will need to figure this out, costing him valuable time. Maybe we could create a bug ticket for someone t figure it out?

src/zsys.c
skaes: If you leave the else branch in, setting any setting of ZSYS_THREAD_NAME_PREFIX would be ignored. And it crashes because you can't strcpy(s,s) in a Mac (and it's illegal on Linux too, but apparently does not crash).

skaes: Personally I think the interface with two env variables is confusing: if the user specifies both variables, which one takes precedence? Also, the underlying implementation in libzmq only uses a single variable to hold the config value, it just uses two different APIs for setting it as either an integer, or a string.

skaes: I'd be happy to create a PR that would use only one env var and calls the correct api depending on whether it has to be a string or not, if you are willing to go this route.

sphaero:
Great to have some more review of this code. We dohave some trouble with OSX and Windows now and then. However this code (multicast) has been tested successfully on OSX. If think it even runs fine on Travis. On what OSX version are you experiencing this trouble?

sphaero:
This also similar to #2111 which we closed because it was working on other machines

src/zsys.c
bluca: Ah yes I see the issue now - regarding the env var, it's just an artifact of how it was implemented in libzmq. It's now in a stable release, so it cannot be removed, but feel free to send a PR to change the implementation.

src/zsock.c
skaes: I think multicast is just not configured on Macs, although they're capable. Only found a very old post: https://blogs.agilefaqs.com/2009/11/08/enabling-multicast-on-your-macos-unix/. Did not want to go down this rabbit hole.

src/zsys.c
skaes: Now I'm confused: I thought zsys_thread_name_prefix_str() and therefore also ZSYS_THREAD_NAME_PREFIX_STR is a draft API and draft APIs can be changed?

If we can't remove that env variable and only use ZSYS_THREAD_NAME_PREFIX as describe in PR comment, I don't know what the PR would contain.

skaes:
@sphaero I'm running on Big Sur.

bluca:
This is how multicast is checked in libzmq selftests:

https://github.com/zeromq/libzmq/blob/master/tests/test_radio_dish.cpp#L307

Can you try if something similar works and send an additional PR?

src/zsys.c
bluca: I was thinking about the old one - the new one is yes, so you can change that. But the behaviour of the old one cannot change - so existing uses cannot break.

skaes: got it

skaes:
@bluca This code is c++ and does not readily compile as part of this project. Sorry, but I currently don't have the time to look into this.

PR #2163, opened 2021-02-09
Fix ntohll/htonll handling

vchuravy:
`ntohll` and `htonll` are not standard and apparently everyone disagrees about it.
`mingw32-w64` is lacking the definition and FreeBSD/NetBSD require a different header.

This hampers cross-compilation of czmq. For the alternative implementations, see https://stackoverflow.com/a/49167531/2442087.

cc: @giordano

### DIFF
--- a/src/zosc.c
+++ b/src/zosc.c
@@ -18,14 +18,22 @@
 @end
 */
 
-#ifdef WIN32
-#include <winsock2.h>           //  needed for ntohll/htonll
-#endif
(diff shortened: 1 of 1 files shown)


bluca:
Add a comment with a link to the source - otherwise it's a license breach, stack overflow is not public domain

bluca:
Windows build is broken

sphaero:
closing as this is already handled by #2180

PR #2168, opened 2021-03-05
Fixed libzmq shared/static cmake targets being both linked against cz…

Mystfit:
```
Problem: Both shared and static LibZMQ targets found from an installed ZeroMQTargets.cmake file are appended to ${MORE_LIBRARIES} which causes double imported symbol issues when linking against the czmq-static target, since it brings in symbols from the shared czmq target as well.

Solution: Split libzmq and libzmq-static targets into ${REQUIRED_LIBRARIES} and ${REQUIRED_LIBRARIES_STATIC} and allow for falling back to PC libzmq if we don't find a cmake config version of libzmq.
```

I'm not sure if this is a fix that should propagated back to the ZProject file?


bluca:
> I'm not sure if this is a fix that should propagated back to the ZProject file?

Yes, otherwise it will just be overwritten

bluca:
also this breaks the build

Mystfit:
> also this breaks the build

Thanks, I was missing the include directories when only building one target. 

I got a bit lost when trying to find where to begin adding this fix to Zproject. Is there a template containing CMake generated code that this fix would be appropriate to add to instead? There's the  ```<use project = "libzmq" />```  entry in project.xml but I need to figure out how that spits out the ```find_package(libzmq REQUIRED)``` block first.

bluca:
https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl

Mystfit:
> https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl

Thanks!

Now I have to scratch my head and ponder how I can modify ```# $(USE.PROJECT) dependency``` to separate the shared and static targets into two variables...

I know that the installed cmake package config name for libzmq is actually ZeroMQ - that divergence is going to be fun to figure out.

Mystfit:
Closing until I figure out how to differentiate between $(USE.PROJECT)'s name being 'libzmq' and ZeroMQ's CMake project name being 'ZeroMQ' so I can move this fix to ZProject.

PR #2177, opened 2021-04-22
fix setsockopt(IP_MULTICAST_IF) on non-Windows systems

yvs2014:
**Problem: 'setsockopt(,IP_MULTICAST_IF,)' fails on linux**

when calling 'zsys_set_ipv4_mcast_address()', for example:
I: 21-04-21 11:47:06 zbeacon: API command=CONFIGURE
I: 21-04-21 11:47:06 zbeacon: interface=eth0 address=192.168.88.27 multicast=239.0.1.2
E: 21-04-21 11:47:06 (UDP) error 'Cannot assign requested address' on zbeacon: setsockopt IP_MULTICAST_IF failed
mcast_test: src/zsys.c:1347: zsys_socket_error: Assertion `false' failed.

or running 'czmq_selftest -t zbeacon':
   zbeacon_ipv4_mcast: SKIPPED - Is IPv4 UDP multicast allowed?
   zbeacon: OK

**Solution:  for example, following linux ip(7)**
>IP_MULTICAST_IF (since Linux 1.2)
>        Set  the  local device for a multicast socket.  The argument for setsockopt(2) is an ip_mreqn or (since Linux 3.5)
>        ip_mreq structure similar to IP_ADD_MEMBERSHIP, or an in_addr structure.  (The kernel determines  which  structure
>        is being passed based on the size passed in optlen.)  For getsockopt(2), the argument is an in_addr structure.

the setsockopt() call needs to be changed from
```c
int if_index;
setsockopt (, IPPROTO_IP, IP_MULTICAST_IF, (char *) &if_index, sizeof (if_index)))
```
to something like this
```c
struct ip_mreqn mreqn;
mreqn.imr_ifindex = if_index;
if (setsockopt (, IPPROTO_IP, IP_MULTICAST_IF, (char *) &mreqn, sizeof (mreqn)))
```

### DIFF
--- a/src/zbeacon.c
+++ b/src/zbeacon.c
@@ -252,7 +252,6 @@ s_self_prepare_udp (self_t *self)
                             self->port_nbr, &hint, &send_to);
                 }
                 assert (rc == 0);
-                assert (rc == 0);
                 if_index = if_nametoindex (name);
 
                 if (self->verbose && !enable_multicast)
@@ -310,6 +309,7 @@ s_self_prepare_udp (self_t *self)
                 zsys_socket_error ("zbeacon: setsockopt IPV6_MULTICAST_IF failed");
         }
         else if(enable_multicast){
+#if defined (__WINDOWS__)
             struct ip_mreq mreq;
             memcpy (&mreq.imr_interface,
                     &(((inaddr_t *) (bind_to->ai_addr))->sin_addr),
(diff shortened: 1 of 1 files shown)


src/zbeacon.c
```diff
@@ -337,6 +337,26 @@ s_self_prepare_udp (self_t *self)
                             (char *) &if_index, sizeof (if_index)))
                 zsys_socket_error (
                   "zbeacon: setsockopt IP_MULTICAST_IF failed");
+#else
```
bluca, line 340: This is not quite right - it's losing IP_ADD_MEMBERSHIP on the sender socket and IP_MULTICAST_IF on the receiver socket. Instead of ifdeffing the whole thing, please restrict it to the IP_MULTICAST_IF calls.

yvs2014: I've not tested it with IP_MULTICAST_IF on the receiver socket and IP_ADD_MEMBERSHIP on the sender.
OK, it needs to be checked on. So, this PR is closed.

PR #2179, opened 2021-05-05
fix zosc leak by moving address to the chunk

sphaero:
Problem: zosc is leaking its address and format string
Solution: keep format on the stack and move address to the chunk like it should

### DIFF
--- a/src/zosc.c
+++ b/src/zosc.c
@@ -150,7 +150,7 @@ zosc_new (const char *address)
     //  Initialize class properties here
     self->format = "";
     self->chunk = zchunk_new(address, strlen(address) + 1);
-    self->address = zchunk_data(self->chunk);
+    self->address = (char *)zchunk_data(self->chunk);
     assert(self->address);
     assert(self->format);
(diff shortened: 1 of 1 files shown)


sphaero:
better wait for tests to finish, my own travis doesn't want to run

src/zosc.c
```diff
@@ -148,11 +148,11 @@ zosc_new (const char *address)
     zosc_t *self = (zosc_t *) zmalloc (sizeof (zosc_t));
     assert (self);
     //  Initialize class properties here
-    self->address = strdup(address);
-    self->format = strdup("");
+    self->format = "";
+    self->chunk = zchunk_new(address, strlen(address) + 1);
+    self->address = zchunk_data(self->chunk);
```
bluca, line 153: src/zosc.c: In function 'zosc_new':
src/zosc.c:153:19: error: pointer targets in assignment differ in signedness [-Werror=pointer-sign]
     self->address = zchunk_data(self->chunk);

PR #2180, opened 2021-05-05
Fix for compilation on BSD and mingw

sphaero:
Problem: As mentioned in #2163 czmq doesn't want to build on BSD and mingw. The provided patch however created new issues. 
Solution: add correct precompiler statements for mentioned platforms

This supersedes #2163 so it can be closed @vchuravy @giordano

### DIFF
--- a/src/zosc.c
+++ b/src/zosc.c
@@ -18,11 +18,17 @@
 @end
 */
 
-#ifdef WIN32
-#include <winsock2.h>           //  needed for ntohll/htonll
-#endif
 #include "czmq_classes.h"
-#if __unix__ && !__APPLE__
+#if defined( __WINDOWS__ )
+#include <winsock2.h>           //  needed for ntohll/htonll
+#elif defined(__UTYPE_FREEBSD) || defined(__UTYPE_NETBSD)
+#include <sys/endian.h>
+#define htonll(x) htobe64(x)
+#define ntohll(x) be64toh(x)
+#elif defined(__MINGW32__)
+#define htonll(x) ((((uint64_t)htonl(x&0xFFFFFFFF)) << 32) + htonl(x >> 32))
+#define ntohll(x) ((((uint64_t)ntohl(x&0xFFFFFFFF)) << 32) + ntohl(x >> 32))
+#elif defined(__UNIX__)
 #include <endian.h>
 #define htonll(x) htobe64(x)
 #define ntohll(x) be64toh(x)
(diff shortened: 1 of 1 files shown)


sphaero:
Better wait for tests to finish, my travis doesn't want to run

src/zosc.c
```diff
@@ -18,11 +18,18 @@
...
 
-#ifdef WIN32
+#include "czmq_classes.h"
+#if defined( __WINDOWS__ )
 #include <winsock2.h>           //  needed for ntohll/htonll
 #endif
-#include "czmq_classes.h"
-#if __unix__ && !__APPLE__
+#if defined(__MINGW32__)
+#define htonll(x) ((((uint64_t)htonl(x&0xFFFFFFFF)) << 32) + htonl(x >> 32))
```
bluca, line 26: C:\projects\czmq\src\zosc.c(74): error C3861: 'htonll': identifier not found [C:\projects\czmq\build\czmq-static.vcxproj]

sphaero: weird the only thing changed was from WIN32 to __WINDOWS__ and moved after czmq_classes.h inclusion. I guess I need a Windows system to check this.

winsock2.h is also included by cmzq_prelude.h :confused:

sphaero: Ok this is not right. cmzq_prelude.h already includes winsock2.h. However it does not provide htonll. However if I include winsock2.h before cmzq_classes.h it does provide it. So somewhere in czmq something is done which results in those methods not being available

giordano: MinGW [`winsock2.h`](https://github.com/mirror/mingw-w64/blob/a7d43b719bc6e9ee050525b4243a9ee1634555d4/mingw-w64-headers/include/winsock2.h) doesn't define `htonll`, which was the point of #2163

sphaero: I think it is caused by this line in winsock2.h

```
#if !defined(NO_EXTRA_HTON_FUNCTIONS) && !defined(__midl) && (defined(INCL_EXTRA_HTON_FUNCTIONS) || NTDDI_VERSION>=NTDDI_WIN8)
```

giordano: Where exactly do you see this line in MinGW `winsock2.h`?

sphaero: I'm talking about winsock2.h from the Windows SDK. This line somehow returns false so the functions are never included.

https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.10240.0/um/WinSock2.h#L1858

If your patch was correct then it should work on mingw now as well

sphaero: It might be caused by zmq.h, if you include winsock2.h before zmq.h it is fixed. I think this belongs in a separate issue. I'll fix it by including winsock2 in zosc.c before anything else like it was.

sphaero:
appveyor is successful now

PR #2183, opened 2021-05-26
OSX is failing on missing endian.h, prevent flooding logs

sphaero:
Problem: OSX is failing to build because of missing endian.h
Solution: add correct include for OSX specifically

additional problem: large zosc message are flooding the logs
solution: only warn for large messages once.


sphaero:
OSX tests are failing on travis because the brew recipe for curl is wrong.

Issue #2188, opened 2021-06-23
OSX builds failing due to curl upstream change


sphaero:
See here https://github.com/curl/curl/pull/6897

It results in failing builds because we need to explicitly set the SSL backend. This is just adding a configure flag however I think it's a platform specific flag. --with-openssl doesn't work on OSX for example. So I think we can't set it in the project.xml.

I'm not sure what's best approach. Perhaps it should be done from zproject?

sphaero:
I have a branch which fixes this problem:

https://github.com/sphaero/czmq/tree/fixosxbuild

CI results: https://travis-ci.com/github/sphaero/czmq/builds/230557025

It seems to work as the linux builds complete. The other OSX failing builds is due to trying to upgrade brew which takes ages

sphaero:
fixed by #2189

PR #2189, opened 2021-06-23
OSX CI build fails due to upstream curl change

sphaero:
Problem: Curl needs a configure option for encryption
Solution: add config option to zproject config


sphaero:
as mentioned in #2188 I'm not sure this is the best cross platform solution. Suggestions welcome

bluca:
if it's a new option, shouldn't we detect the version and use it conditionally? Otherwise backward compat will break, no?

sphaero:
I think this is only used in the ci_build.sh script which does a git clone of the master repo: https://github.com/zeromq/czmq/blob/b30fc4288e9241dc4d7cd562327c9db0a6c53c4f/ci_build.sh#L346

So there's no versioning involved currently. Zproject does have some versioning params but I don't think they are used either.

bluca:
Ah yes of course, good point

PR #2193, opened 2021-07-16
Added stream socket test

sphaero:
Problem: There's no STREAM socket test
Solution: add a test similar to libzmq's

### DIFF
--- a/src/zsock.c
+++ b/src/zsock.c
@@ -2203,11 +2203,15 @@ zsock_test (bool verbose)
     zmsg_destroy (&msg);
 
 #ifdef ZMQ_STREAM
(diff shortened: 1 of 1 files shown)


src/zsock.c
```diff
@@ -2202,6 +2202,104 @@ zsock_test (bool verbose)
     zframe_destroy (&frame);
     zmsg_destroy (&msg);
 
+#ifdef ZMQ_STREAM
+    zsock_t *streamrecv = zsock_new_stream("@tcp://127.0.0.1:1234");
```
bluca, line 2206: please do not hard-code the port, otherwise the test will randomly fail - use `*` and retrieve it

sphaero: done

PR #2195, opened 2021-08-20
Adding zoscdump utility

sphaero:
This is not so much as fixing a problem but rather adding functionality. 

As a convenience this PR adds the `zoscdump` utility which can receive OSC messages from TCP or UDP transports and prints them to stdout similar to liblo's oscdump. This is often useful for debugging other software and appliances.


sphaero:
This is a draft PR as I'm not sure this is wanted here. So comments welcomed.

PR #2206, opened 2022-03-03
problem: corrupted zframes called in zmsg_decode create a crash instead of returning NULL

stvales:
Solution:
- in zframe, use freen instead of zframe_destroy when frame init fails (heap exhaustion which may be due to message corruption) to avoid a bad memory access on the still unallocated self->zmsg
- in zmsg, break and return NULL instead of asserting after trying to initiate the zmsg from the decoded frame if the frame is NULL

### DIFF
--- a/src/zmsg.c
+++ b/src/zmsg.c
@@ -701,7 +701,10 @@ zmsg_decode (zframe_t *frame)
             break;
         }
         zframe_t *decoded = zframe_new (source, frame_size);
-        assert (decoded);
+        if (!decoded) {
+            zmsg_destroy (&self);
+            break;
+        }
         zmsg_append (self, &decoded);
         source += frame_size;
     }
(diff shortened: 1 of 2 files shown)


src/zmsg.c
```diff
@@ -701,7 +701,10 @@ zmsg_decode (zframe_t *frame)
             break;
         }
         zframe_t *decoded = zframe_new (source, frame_size);
-        assert (decoded);
+        if (!decoded) {
```
bluca, line 704: in what cases can this happen?

stvales: This happens if frame_size causes a heap exhaustion which makes zframe_new return NULL (see zframe.c).

bluca: we don't really support running when there's no memory available, it will break in millions of other ways

stvales: We found this one by trying to decode data corresponding to a PNG image. This first bytes normally determine the size of the zframe but in this case, it created a heap exhaustion due to the decoded size. It's not strictly heap exhaustion but it may happen when data to be decoded is corrupted or not an encoded zframe.

PR #2223, opened 2022-05-17
Problem: zdir_flatten pushes subdirs at the end causing problems with patches computation in zdir_diff

stvales:
Problem: zdir_flatten pushes subdirs at the end causing problems with patches computation in zdir_diff
For example, when a file is added in the root directory and alphabetically placed between two subdirs, zdir_diff generates false create/delete patches for all the files in the subdirs.

Solution: sort the whole flat list generated by zdir_flatten after all recursions are over.

### DIFF
--- a/src/zdir.c
+++ b/src/zdir.c
@@ -341,7 +341,7 @@ zdir_flatten (zdir_t *self)
         zlist_append (sorted, files[i]);
     zlist_sort (sorted, s_file_compare);
     for (size_t i = 0; i < self->count; i++)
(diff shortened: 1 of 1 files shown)


bluca:
```
C:\projects\czmq\src\zdir.c(344): error C2440: '=': cannot convert from 'void *' to 'zfile_t *' [C:\projects\czmq\build\czmq.vcxproj]
  C:\projects\czmq\src\zdir.c(344): note: Conversion from 'void*' to pointer to non-'void' requires an explicit cast
```

PR #2228, opened 2022-08-04
[CMake] Check if libzmq is already found

NouberNou:
This allows a CMake project that includes `libzmq` and `czmq` as source to define how to find `libzmq` if the `find_package` module is inadequate.

These inadequacies arise due to situations where `find_package` is not optimal, such as a source/git submodule distribution of all dependencies rather than system installs of dependencies (which can become unwieldy in CI/CD).


CMakeLists.txt
```diff
@@ -130,16 +130,18 @@ set(OPTIONAL_LIBRARIES_STATIC)
...
 # LIBZMQ dependency
 ########################################################################
-find_package(libzmq REQUIRED)
+if (NOT libzmq_FOUND)
+    find_package(libzmq REQUIRED)
+endif()
 IF (libzmq_FOUND)
     include_directories(${libzmq_INCLUDE_DIRS})
     list(APPEND MORE_LIBRARIES ${libzmq_LIBRARIES})
-  IF (PC_LIBZMQ_FOUND)
```
bluca, line 137: no need to change this whitespace

CMakeLists.txt
```diff
@@ -130,16 +130,18 @@ set(OPTIONAL_LIBRARIES_STATIC)
 ########################################################################
 # LIBZMQ dependency
 ########################################################################
-find_package(libzmq REQUIRED)
+if (NOT libzmq_FOUND)
```
bluca, line 133: add it here or it will be lost on the next iteration https://github.com/zeromq/zproject/blob/master/zproject_cmake.gsl

NouberNou: Roger, what is the best way to test a build of zproject against czmq to make sure the output is the same? The change is relatively simple in the gsl file and if you guys are open to a PR on that _without_ me locally testing I'd prefer it (for the sake of keeping this simple on my end).

bluca: it's a simple enough change and there's a CI if it breaks something, so just send it

NouberNou: OK, I'll create the PR in that project and link it here to close this one out, thanks.

NouberNou:
See zproject PR here for long-term patch: https://github.com/zeromq/zproject/pull/1298

PR #2230, opened 2022-08-21
zframe.c/print: reduce runtime cost

jeffdegott:
reduce runtime cost when checking if the frame is binary or not,
in zframe_print_n()  & zframe_fprint()


paddor:
@FrancoisRegisDegott-eaton next time please supply a more descriptive commit message.

PR #2248, opened 2022-11-01
zproxy: Add heartbeat command

### LINKED ISSUE #2244: Lingering broken TCP connections when using zproxy
Hello!

We are using czmq to create a broker on a Linux server that handles quite a few devices over quite long periods of time. We set up a "pubsub" proxy to handle some requests from the devices as such:

```c
zactor_t *
create_pubsub_proxy (int in_port, int out_port) {
    zactor_t *proxy=NULL;

    proxy = zactor_new (zproxy, NULL); assert (proxy);
    
    // Set the frontend (incoming messages)
    zstr_sendm (proxy, "FRONTEND");
    zstr_sendm (proxy, "XSUB");
    zstr_sendf(proxy, "tcp://*:%d", in_port);
    zsock_wait (proxy);
    
    // Set backend, (outgoing messages)
    zstr_sendm (proxy, "BACKEND");
    zstr_sendm (proxy, "XPUB");
    zstr_sendf(proxy, "tcp://*:%d", out_port);
    zsock_wait (proxy);
    
    return proxy;
}
```
We notice that the resource allocation (memory) of the broker is constantly growing and it tracks with the amount of open file descriptors. I can reproduce (on GitHub master of libczmq) lingering TCP file descriptors by issuing the following command on a device:

```bash
$ (ip link set eth0 down; sleep 3; reboot) &
```
I have looked for similar issues and found https://github.com/zeromq/libzmq/issues/1453 which talks about adding TCP_KEEPALIVE to fix this issue. But there is no way for me to set keepalive on the proxy sockets.

When I added the following diff to libczmq code:

```diff
diff --git a/src/zproxy.c b/src/zproxy.c
index 90dd5a8f..fc0fd8d6 100644
--- a/src/zproxy.c
+++ b/src/zproxy.c
@@ -105,6 +105,11 @@ s_self_create_socket (self_t *self, char *type_name, char *endpoints, proxy_sock
     }
     zsock_t *sock = zsock_new (index);
     if (sock) {
+        printf("setting sockopts!\n");
+        zsock_set_tcp_keepalive(sock, 1);
+        zsock_set_tcp_keepalive_idle(sock, 30);
+        zsock_set_tcp_keepalive_intvl(sock, 3);
+        zsock_set_tcp_keepalive_cnt(sock, 10);
 #if (ZMQ_VERSION_MAJOR == 4)
         if (self->domain [selected_socket]) {
             // Apply authentication domain
```
Then the lingering socket went a way after ~2 minutes.

Is there something else I can tweak? Or should I try to add a nicer way to add keepalive to zproxy sockets?

jonasdn:
In order to clean out broken TCP connections with the proxy.

Example usage:
```c
    proxy = zactor_new (zproxy, NULL); assert (proxy);

    // Set the frontend (incoming messages)
    char endpoint[64] =  { 0, };
    sprintf(endpoint, "tcp://*:%d", in_port);
    zstr_sendx (proxy, "FRONTEND", "XSUB", endpoint,
        "HEARTBEAT_IVL", "300000", "HEARTBEAT_TIMEOUT", "30000",
        "HEARTBEAT_TTL", "0", NULL);
    zsock_wait (proxy);
```

Fixes #2244

### DIFF
--- a/src/zproxy.c
+++ b/src/zproxy.c
@@ -32,6 +32,12 @@ typedef enum proxy_socket {
 }
 proxy_socket;
 
+typedef struct {
+    int ivl;
+    int timeout;
+    int ttl;
+} zproxy_heartbeat_t;
+
 #define AUTH_NONE  0
 #define AUTH_PLAIN 1
 #define AUTH_CURVE 2
@@ -87,7 +93,7 @@ s_self_destroy (self_t **self_p)
 }
 
 static zsock_t *
-s_self_create_socket (self_t *self, char *type_name, char *endpoints, proxy_socket selected_socket)
+s_self_create_socket (self_t *self, char *type_name, char *endpoints, zproxy_heartbeat_t *heartbeat, proxy_socket selected_socket)
 {
     //  This array matches ZMQ_XXX type definitions
     assert (ZMQ_PAIR == 0);
@@ -126,6 +132,15 @@ s_self_create_socket (self_t *self, char *type_name, char *endpoints, proxy_sock
 
             // Enable curve authentication
             zsock_set_curve_server (sock, 1);
+        }
+        if (heartbeat->ivl != 0) {
+            if (self->verbose) {
+                zsys_info ("zproxy - setting heartbeat ivl=%d, timeout=%d, ttl=%d",
+                    heartbeat->ivl, heartbeat->timeout, heartbeat->ttl);
+            }
+            zsock_set_heartbeat_ivl (sock, heartbeat->ivl);
+            zsock_set_heartbeat_timeout (sock, heartbeat->timeout);
+            zsock_set_heartbeat_ttl (sock, heartbeat->ttl);
         }
 #endif
         if (zsock_attach (sock, endpoints, true)) {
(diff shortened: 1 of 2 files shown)


jonasdn:
Could something like this, with some kind of hearbeat-command to zproxy be considered? Or is there perhaps a better approach for this problem (#2244) that I am missing?

Thank you!

src/zproxy.c
```diff
@@ -191,12 +203,23 @@ s_self_configure (self_t *self, zsock_t **sock_p, zmsg_t *request, proxy_socket
     assert (type_name);
     char *endpoints = zmsg_popstr (request);
     assert (endpoints);
+
+    zproxy_heartbeat_t zproxy_heartbeat = { 0, };
+    char *heartbeat = zmsg_popstr (request);
+    if (heartbeat && strncmp (heartbeat, "HEARTBEAT", strlen ("HEARTBEAT")) == 0) {
```
bluca, line 209: don't cram it like this, use separate message parts for each parameter instead

src/zproxy.c
```diff
@@ -127,6 +133,12 @@ s_self_create_socket (self_t *self, char *type_name, char *endpoints, proxy_sock
             // Enable curve authentication
             zsock_set_curve_server (sock, 1);
         }
+
+        if (heartbeat) {
```
bluca, line 137: this check is always true regardless of the content, which means the parameters are always set regardless of whether they are passed or not, which is not right

jonasdn:
> docs and unit test changes are missing

Thanks! I added some, let me know if I should refine them!

src/zproxy.c
jonasdn: You are right, thank you, I changed to checking the `ivl` field, if that is zero no other option matters.

src/zproxy.c
jonasdn: Thanks! I changed to multiple message parts!

src/zproxy.c
```diff
@@ -183,6 +198,32 @@ s_self_selected_socket (zmsg_t *request)
     return socket;
 }
 
+static int
+s_get_int_value (zmsg_t *request)
+{
+    char *value = zmsg_popstr (request);
```
bluca, line 204: strings from zmsg_popstr need to be freed

jonasdn: Thank you!

PR #2255, opened 2022-12-05
Problem: Sync with zproject (Android build fails if linkname is different than prefix).

stephan57160:
Solution: Regenerate CZMQ with this commit.

Note:

There are many modified files, because of:

    libraries.put("libsystemd", true);          # buggy

being replaced by:

    libraries.put("systemd", true);             # correct form

in each generated .java files (and there are many classes in CZMQ ...)

### DIFF
--- a/bindings/jni/czmq-jni-native/build.gradle
+++ b/bindings/jni/czmq-jni-native/build.gradle
@@ -50,7 +50,7 @@ task copyLibs(type: Copy) {
             include '*uuid*.dll'
             include 'libsystemd.so'
             include 'libsystemd.dylib'
-            include '*libsystemd*.dll'
+            include '*systemd*.dll'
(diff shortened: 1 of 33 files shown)


stephan57160:
Had a look to failing tests.
I don't think the failures come from this fix.
Most of them fail when updating Ubuntu packages:

    The following packages have unmet dependencies:
     libunwind-14-dev : Breaks: libunwind-dev but 1.3.2-2build2 is to be installed
     libzmq5 : Depends: libpgm-5.2-0 (>= 5.1.116~dfsg) but it is not installable
    E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
    Error: Process completed with exit code 100.

Could be an idea to relaunch the tests, later today, if possible...

Issue #2144, opened 2021-01-04
zhashx_dup() should use previous hashes comparator and key functions?


chu11:
Recently used `zhashx_dup()` to copy a hash, leading to an assert in `zhashx_insert()` while in the middle of the copy iteration in `zhashx_dup()`.

Eventually, I determined the issue was that we use an alternate setting of the key hasher, comparator, duplicator, and destructor.  However in `zhashx_dup()` it does not appear to copy those over.  It only copies over the item duplicator/destructor.

```
    zhashx_t *copy = zhashx_new ();
    if (copy) {
        copy->destructor = self->destructor;
        copy->duplicator = self->duplicator;
```

I think it should be using the same hasher/key functions across the board?

bluca:
Can you send a PR to fix it, please?

chu11:
@bluca sure

Issue #1193, opened 2015-11-28
Problem: CZMQ codebase is filled with useless OOM checks


hintjens:
This code was added at a point, to make CZMQ continue safely if a process ran out of memory.

It makes the codebase more complex in practically every method. 

Actually testing out-of-memory conditions shows that these checks do not work: libzmq and other libraries randomly die when the heap is exhausted.

I'd like to replace all these checks with asserts. Poorly-sized memory limits are not a recoverable problem IMO.

Solution: rip out OOM error handling and replace with asserts.

Asmod4n:
The problem with asserts is, they can be turned off. So in some cases you have to check for NULL or do something else like abort(); maybe.

hintjens:
I don't think we ever turn off the asserts in libzmq or CZMQ. Anyone doing
that in production code gets what they deserve...

Asmod4n:
Debian does this with release packages by default I believe, they are turned on for debug packages. I could beer wrong tho.

bluca:
I can confirm in Debian both libzmq and CZMQ are built with the default options, so asserts are in. And rightly so :-)

https://anonscm.debian.org/cgit/collab-maint/czmq.git/tree/debian/rules
https://anonscm.debian.org/cgit/collab-maint/zeromq3.git/tree/debian/rules

Asmod4n:
yay :)

PR #2266, opened 2023-05-30
Update zlist/zlistx header comments to indicate memory exhaustion not possible error

chu11:
I noticed a few comments in the headers for zlist/zlistx that mention errors are returned on out of memory conditions.  I believe those errors are no longer possible, as czmq is built with asserts to bail if any internal memory allocation fails.  Apologies if I'm wrong about this.

For `zlist`, I just changed the example error of "out of memory" to "invalid input", which is a real possible error.  For `zlistx`, I just removed the offending text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5138,13 +5138,13 @@ This is the class interface:
         zlist_item (zlist_t *self);
     
     //  Append an item to the end of the list, return 0 if OK or -1 if this
-    //  failed for some reason (out of memory). Note that if a duplicator has
+    //  failed for some reason (invalid input). Note that if a duplicator has
     //  been set, this method will also duplicate the item.
     CZMQ_EXPORT int
         zlist_append (zlist_t *self, void *item);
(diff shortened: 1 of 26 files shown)


sphaero:
I'm not sure if you are correct. Often an assert does nothing when you do a release build. So in that case a memory exhaustion can happen.

chu11:
generally speaking you are correct, but it appears czmq builds them in.  see:

https://github.com/zeromq/czmq/issues/1193

unless czmq has changed since that thread.

I can minimally confirm to have hit some asserts in some czmq classes on some other random bugs using the czmq built by redhat.

https://github.com/zeromq/czmq/issues/2144

sphaero:
Yeah, wouldn't wanna do without these asserts. Sorry for the delay in merging this PR. If you are a frequent user of czmq or other 0mq projects we could use some help handling these PR's.

PR #2269, opened 2023-07-10
Problem: zdir_list omits empty subdirectories

Felcanio:
Solution: zdir_list_paths return a char* list of all the files and subdirectories of zdir


bluca:
Please see the top of `include/zdir.h`, this change will be overwritten and lost on the next regen otherwise

bluca:
You need to regenerate using gsl, see the README

bluca:
some changes are still missing, see CI job result

PR #2275, opened 2023-12-09
fixup invalid XML entities in document

flier:
Problem: Invalid XML entities in document

Solution: use XML entity or warp document in `<![CDATA[]>`


sphaero:
I'm not sure if we want this since this makes the generated api docs unreadable from source.

Can you elaborate what the problem is?

sphaero:
No more progress so closing...

PR #2282, opened 2024-04-02
Fixed issue 2278, loop->terminated == true when restarting a loop

ihatechoosingusernames:
Problem: Issue 2278; loop->terminated == true when restarting a loop allowing timer callbacks to update loop->timers in place.

Solution: Set loop->terminated = false at beginning of zloop_start. Add test to both demonstrate the issue and test the fix.

### DIFF
--- a/src/zloop.c
+++ b/src/zloop.c
@@ -709,6 +709,7 @@ zloop_start (zloop_t *self)
 {
     assert (self);
     int rc = 0;
+    self->terminated = false;
 
     //  Main reactor loop
     while (!zsys_interrupted || self->nonstop) {


sphaero:
It seems to fail on macOS?
```
make[3]: Circular src/selftest-ro <- src/selftest-ro dependency dropped.
/bin/sh ./libtool --mode=execute ./src/czmq_selftest
Assertion failed: (timer_end_called_1 == false), function zloop_test, file zloop.c, line 1027.
Running czmq selftests...
make[3]: *** [check-local] Abort trap: 6
 * zgossip_msg: OK
 * zactor: OK
 * zarmour: OK
 * zcert: OK
make[2]: *** [check-am] Error 2
 * zcertstore: OK
 * zchunk: OK
 * zclock: OK
 * zconfig: OK
 * zdigest: OK
 * zdir: OK
 * zdir_patch: OK
 * zfile: OK
 * zframe: OK
 * zhash: OK
 * zhashx: OK
 * ziflist: OK
 * zlist: OK
 * zlistx: OK
make[1]: *** [check-recursive] Error 1
make: *** [distcheck] Error 1
 * zloop: 
Error: Process completed with exit code 2.
```

PR #2283, opened 2024-04-05
Problem: Out of date zproject

BHare1985:
Solution: Run GSL on the project.xml

These changes are potentially superfluous, but since I am running `tstgenbld.sh`, I might as well update most projects to the latest versions while learning zproject.

There were two misspellings that was fixed on czmq that came from zproject generation (and thus regenerating them would bring back the misspellings) via https://github.com/zeromq/czmq/pull/2273. I fixed the one related to the API and I ommitted the one that came from zproject itself, as I will make a pull request to zproject to fix it.


sphaero:
thanks

PR #2288, opened 2024-06-19
Create an OSC message from a string

sphaero:
Problem: it's not easy to create a message simply by describing it like how a message can be printed as a string
Solution: Add zosc_fromstring method enabling the creation of message from a string

Also added test code.

A message can be constructed as follows:
```c
zosc_t* msg = zosc_fromstring("/stringmsg ihfdscF 32 64 1.100000 3.140000 hello q");
```

PR #2293, opened 2024-08-02
Regenerate from zproject

benjdero:


sphaero:
we might need to take a look at why  zhttp_client_actor  is failing valgrind tests

benjdero:
It's failing since #2288, so can you accept this PR as it has nothing to do with it?

sphaero:
yes, it's weird that change didn't touch anything from zhttp. Must be the env? Merging this one

PR #2294, opened 2024-08-05
Problem: zhttp_client does not support PUT or PATCH

stephenprocter:
Solution: Add support for PUT and PATCH methods.


sphaero:
Thnx!

PR #2295, opened 2024-08-06
Problem: zhttp_client does not support authentication

stephenprocter:
Solution: Add functions to zhttp_request to set username and password, and send these to the client.


sphaero:
you are missing some zproject regenerated files. can you commit those as well?

Issue #2299, opened 2024-08-16
zbeacon test Segmentation fault


calvin2021y:
If there no interface or only abnormal interface, linux:

```sh
Thread 4 "ZACTOR" received signal SIGSEGV, Segmentation fault.
[Switching to LWP 69824]
0x00000000005951f7 in freeaddrinfo (p=0x0) at src/network/freeaddrinfo.c:10
10	src/network/freeaddrinfo.c: No such file or directory.
(gdb) bt
#0  0x00000000005951f7 in freeaddrinfo (p=0x0) at src/network/freeaddrinfo.c:10
#1  0x0000000000396b34 in s_self_prepare_udp (self=0x7f0ede01da40) at czmq/src/zbeacon.c:372
#2  0x000000000039805f in s_self_configure (self=0x7f0ede01da40, port_nbr=9999) at czmq/src/zbeacon.c:386
#3  0x0000000000395376 in s_self_handle_pipe (self=0x7f0ede01da40) at czmq/src/zbeacon.c:414
#4  0x00000000003950cf in zbeacon (pipe=0x7f0ede074080, args=0x0) at czmq/src/zbeacon.c:568
#5  0x00000000003de387 in s_thread_shim (args=0x7f0ede074470) at czmq/src/zactor.c:68
#6  0x00000000005a79ee in start (p=0x7f0ede012af0) at src/thread/pthread_create.c:207
#7  0x00000000005a9621 in __clone () at src/thread/x86_64/clone.s:22
```

calvin2021y:
This error only occurs when using musl; there are no issues with glibc.

calvin2021y:
this patch fix the problem

```diff
diff --git a/src/zbeacon.c b/src/zbeacon.c
index ee897406..ccff27a5 100644
--- a/src/zbeacon.c
+++ b/src/zbeacon.c
@@ -369,8 +369,12 @@ s_self_prepare_udp (self_t *self)
         self->udpsock_send = INVALID_SOCKET;
     }
 
-    freeaddrinfo (bind_to);
-    freeaddrinfo (send_to);
+       if( bind_to != NULL ) {
+           freeaddrinfo (bind_to);
+       }
+       if( send_to !=NULL && bind_to != send_to ) {
+           freeaddrinfo (send_to);
+       }
 }
```

sphaero:
Isn't it that if send_to and bind_to are NULL we have some error state?

calvin2021y:
[...]

PR #2303, opened 2024-09-12
fix zbeacon test segmentation fault

calvin2021y:
# Pull Request Notice
https://github.com/zeromq/czmq/issues/2299


calvin2021y:
the error here is new and not sure how to fix it.  will try more.

PR #2327, opened 2025-11-18
Fix to make czmq/zyre build in MSVC 2022 when C version is <23

zedalaye:
Problem: cannot build zyre with MSVC 2022 when C is < v23

Solution: use a better way to detect when stdbool.h is available


sphaero:
Thnx

PR #2328, opened 2026-02-19
Fix technical undefined behavior with snprintf

jdavidberger:
To be clear there is no input / data path that could cause a buffer overflow; however there are various instrumentation / dynamic checks for this sort of thing that this code triggers (along with the general issue with it being UB).

Specifically [this snprintf implementation](http://refspecs.linux-foundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/libc---snprintf-chk-1.html) aborts here.


sphaero:
Thanks

PR #2329, opened 2026-04-13
Fix zlistx_sort edge case

Nistenf:
The zlistx_sort comb sort implementation was missing the minimum gap value check. This causes any array which needs more than one gap==1 passes to be sorted to end up using a gap value of 0, comparing each item to itself and returning early without finishing the sort.
This PR adds the missing check and a (previously failing) test case.


sphaero:
Thanks!

Issue #2330, opened 2026-05-18
build fails on Ubuntu26.04


michelpromonet:
running
```
sudo apt install libzmq3-dev
cmake .
make
```
fails with
```
[  1%] Building C object CMakeFiles/czmq_objects.dir/src/zactor.c.o
/czmq/src/zactor.c: In function ‘zactor_new’:
/czmq/src/zactor.c:129:5: error: implicit declaration of function ‘pthread_setname_np’; did you mean ‘pthread_setcanceltype’? [-Wimplicit-function-declaration]
  129 |     pthread_setname_np(thread, "ZACTOR");
      |     ^~~~~~~~~~~~~~~~~~
      |     pthread_setcanceltype
make[2]: *** [CMakeFiles/czmq_objects.dir/build.make:79: CMakeFiles/czmq_objects.dir/src/zactor.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:144: CMakeFiles/czmq_objects.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

mpromonet:
It seems same problems than https://github.com/zeromq/czmq/issues/2305

Issue #2305, opened 2024-10-04
 implicit declaration of function ‘pthread_setname_np’


sphaero:
Apparently we need to set:

`#define _GNU_SOURCE `

before we #include <pthread.h>

Not sure how portable that is. We could also revert it.

oko256:
Do you know if there's any downside to just always define `_GNU_SOURCE`? I guess platforms that don't support it would just ignore it?

I noticed that _GNU_SOURCE is defined globally by zproject if building via autotools, but such definition is not added to CMake variant of the build files and on many occasions I would really like to use CMake instead.

oko256:
I kinda agree that it's probably not worth the hassle and it would just make sense to drop the thread naming completely.

PR #2331, opened 2026-05-18
Add definition for GNU source on Linux

michelpromonet:
Problem:
```
/czmq/src/zactor.c: In function ‘zactor_new’:
/czmq/src/zactor.c:129:5: error: implicit declaration of function ‘pthread_setname_np’; did you mean ‘pthread_setcanceltype’? [-Wimplicit-function-declaration]
```
see https://github.com/zeromq/czmq/issues/2330

Solution: define _GNU_SOURCE for linux


mpromonet:
It seems similar to https://github.com/zeromq/czmq/issues/2305

